PR #4, opened 2015-07-30
Docker client refactor

ibuildthecloud:
The change adds the concept of a ClientFactory for the docker service implementation.  This will allow the client construction to be more extensible and dynamic.  Such a thing is a needed for RancherOS in that we dynamically switch back and forth between two Docker daemons from the same project.

Additionally this code makes the default project listener exported so that somebody can external use it.  This is also used in RancherOS as we would like to use the default listener in addition to another listener we provide to add more capabilities.


cpuguy83:
Trying to run this, seems as though the tls options are not working, either from env or when setting them manually through the flags.
It's trying to connect over non-tls.

ibuildthecloud:
Thanks @cpuguy83, I'll test locally again.  I very well could have messed up something in the factor.

ibuildthecloud:
@cpuguy83 The CLI argument parsing was messed up.  Its working now.

PR #7, opened 2015-08-06
Make it unambiguous that libcompose is a library

bfirsh:
We need to make sure that users don't accidentally pick this up and start trying to use it instead of the main version of Compose, otherwise they'll run into incompatibilities and it'll be a pain to support.

This may change in the future, but for now we should be 100% clear about what it is or users will get confused. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # libcompose
 
-An official implementation of Docker Compose in Go made available as a library.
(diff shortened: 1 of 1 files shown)


aanand:
LGTM

ibuildthecloud:
LGTM

PR #8, opened 2015-08-06
docker: support relative paths for volumes

mssola:
Consider the following yaml file:

```yaml
test:
  image: opensuse:13.2
  volumes:
    - foo:/foo
  command: sh -c 'echo Hello World'
```

docker-compose can handle the above configuration file. However this
implementation fails because it cannot handle the relative path "foo". This
commit fixes this by converting relative paths to absolute paths when
converting the given config file.

Moreover, this commit also merges two for loops that were dealing with volumes.

Signed-off-by: Miquel Sabaté <msabate@suse.com>

### DIFF
--- a/docker/convert.go
+++ b/docker/convert.go
@@ -31,8 +31,8 @@ func isVolume(s string) bool {
 }
 
 // ConvertToAPI converts a service configuration to a docker API container configuration.
-func ConvertToAPI(c *project.ServiceConfig) (*dockerclient.ContainerConfig, error) {
-	config, hostConfig, err := Convert(c)
+func ConvertToAPI(s *Service) (*dockerclient.ContainerConfig, error) {
+	config, hostConfig, err := Convert(s.serviceConfig, s.context)
 	if err != nil {
 		return nil, err
 	}
@@ -52,12 +52,15 @@ func ConvertToAPI(c *project.ServiceConfig) (*dockerclient.ContainerConfig, erro
 }
 
 // Convert converts a service configuration to an docker inner representation (using runconfig structures)
-func Convert(c *project.ServiceConfig) (*runconfig.Config, *runconfig.HostConfig, error) {
-	vs := Filter(c.Volumes, isVolume)
+func Convert(c *project.ServiceConfig, ctx *Context) (*runconfig.Config, *runconfig.HostConfig, error) {
+	volumes := make(map[string]struct{}, len(c.Volumes))
+	for k, v := range c.Volumes {
+		vol := ctx.ResourceLookup.ResolvePath(v, ctx.ComposeFile)
 
-	volumes := make(map[string]struct{}, len(vs))
-	for _, v := range vs {
-		volumes[v] = struct{}{}
+		c.Volumes[k] = vol
+		if isVolume(vol) {
+			volumes[vol] = struct{}{}
+		}
 	}
 
 	ports, binding, err := nat.ParsePortSpecs(c.Ports)
--- a/docker/convert_test.go
+++ b/docker/convert_test.go
@@ -1,10 +1,13 @@
 package docker
 
 import (
+	"path/filepath"
+	"testing"
+
+	"github.com/docker/libcompose/lookup"
 	"github.com/docker/libcompose/project"
 	shlex "github.com/flynn/go-shlex"
 	"github.com/stretchr/testify/assert"
-	"testing"
 )
 
 func TestParseCommand(t *testing.T) {
@@ -15,10 +18,16 @@ func TestParseCommand(t *testing.T) {
 }
 
 func TestParseBindsAndVolumes(t *testing.T) {
+	ctx := &Context{}
+	ctx.ComposeFile = "foo/docker-compose.yml"
+	ctx.ResourceLookup = &lookup.FileConfigLookup{}
+
+	abs, err := filepath.Abs(".")
+	assert.Nil(t, err)
 	cfg, hostCfg, err := Convert(&project.ServiceConfig{
-		Volumes: []string{"/foo", "/home:/home", "/bar/baz", "/usr/lib:/usr/lib:ro"},
-	})
+		Volumes: []string{"/foo", "/home:/home", "/bar/baz", ".:/home", "/usr/lib:/usr/lib:ro"},
+	}, ctx)
 	assert.Nil(t, err)
 	assert.Equal(t, map[string]struct{}{"/foo": {}, "/bar/baz": {}}, cfg.Volumes)
-	assert.Equal(t, []string{"/home:/home", "/usr/lib:/usr/lib:ro"}, hostCfg.Binds)
+	assert.Equal(t, []string{"/home:/home", abs + "/foo:/home", "/usr/lib:/usr/lib:ro"}, hostCfg.Binds)
 }
--- a/docker/container.go
+++ b/docker/container.go
@@ -212,7 +212,7 @@ func (c *Container) OutOfSync() (bool, error) {
 }
 
 func (c *Container) createContainer(imageName string) (*dockerclient.Container, error) {
-	config, err := ConvertToAPI(c.service.serviceConfig)
+	config, err := ConvertToAPI(c.service)
 	if err != nil {
 		return nil, err
 	}
--- a/docker/project.go
+++ b/docker/project.go
@@ -9,8 +9,8 @@ import (
 
 // NewProject creates a Project with the specified context.
 func NewProject(context *Context) (*project.Project, error) {
-	if context.ConfigLookup == nil {
-		context.ConfigLookup = &lookup.FileConfigLookup{}
+	if context.ResourceLookup == nil {
+		context.ResourceLookup = &lookup.FileConfigLookup{}
 	}
 
 	if context.EnvironmentLookup == nil {
--- a/integration/basic_test.go
+++ b/integration/basic_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"testing"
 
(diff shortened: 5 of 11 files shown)


docker/convert.go
```diff
@@ -29,6 +30,21 @@ func isVolume(s string) bool {
...
 
+func toAbs(s string) string {
+	vs := strings.SplitN(s, ":", 2)
+	if len(vs) != 2 {
+		return s
+	}
+	if !filepath.IsAbs(vs[0]) {
+		if path, err := filepath.Abs(vs[0]); err != nil {
+			return s
+		} else {
```
cpuguy83: This is a redundant else.

```
if path,err := .... {
    return s
}
vs[0] = path

File: docker/convert_test.go [hunk 2]
```diff
@@ -16,9 +19,11 @@ func TestParseCommand(t *testing.T) {
 
 func TestParseBindsAndVolumes(t *testing.T) {
 	cfg, hostCfg, err := Convert(&project.ServiceConfig{
-		Volumes: []string{"/foo", "/home:/home", "/bar/baz", "/usr/lib:/usr/lib:ro"},
+		Volumes: []string{"/foo", "/home:/home", "/bar/baz", ".:/home", "/usr/lib:/usr/lib:ro"},
 	})
 	assert.Nil(t, err)
 	assert.Equal(t, map[string]struct{}{"/foo": {}, "/bar/baz": {}}, cfg.Volumes)
-	assert.Equal(t, []string{"/home:/home", "/usr/lib:/usr/lib:ro"}, hostCfg.Binds)
+	abs, _ := filepath.Abs(".")
```
cpuguy83: Please check errors

docker/convert.go
```diff
@@ -29,6 +30,21 @@ func isVolume(s string) bool {
 	return !isBind(s)
 }
 
+func toAbs(s string) string {
```
cpuguy83: Even though this is a not exported, can you add a doc string so others who come across know the what this function is specifically for?

ibuildthecloud:
Good catch, thanks for finding this.

Part of this project is possibly breaking some compatibility with the current Docker compose for the greater good.  This syntax conflicts with volume drivers.  I would prefer that we require the path to start with `./`.  @cpuguy83 has this discussion come up before?

cpuguy83:
@ibuildthecloud It has not.
TBH I like this syntax in compose... but you're right it does conflict with setting volume names.

If we want to be able to set a volume name I think having a config spec for it is better than following the docker CLI explicitly since CLI's have limitations that we can overcome here.

Definitely worthy of discussion on the right way to solve this rather than just copying py-compose.

cpuguy83:
Actually, I've had several questions from people with regard to how compose plans to handle the new volumes API (plugins and the top-level api commands).

aanand:
Compose only performs path expansion if you haven't specified a volume driver.

ibuildthecloud:
@aanand The problem with that approach is that once volumes become a top level entity (`docker volume create`), we like to experiment with controlling volumes from compose.  For example:

```yaml
myvolume:
  type: volume
  driver: vfs

myapp:
  volumes:
  - myvolume:/foo
```

In this situation you wouldn't have a volume driver in the service definition.

docker/convert.go
mssola: Your suggestion didn't compile, so I re-formatted the code in a cleaner way ;)

cpuguy83:
It's also not really valid since you can name local volumes as well.

aanand:
How does the docker CLI distinguish between a volume path and a named volume? Is it whether or not it begins with `/`? If so, I'm happy to do the same with Compose, but extend it to "begins with `/` or `.`".

This will be a backwards-incompatible change, of course - `volumes: ["data:/data"]` will now parse differently - so perhaps we should just put a deprecation warning into the Compose 1.4 release for when we encounter relative paths without dots or slashes, then make the actual change in 1.5.

aanand:
I've created an issue about this in Compose: https://github.com/docker/compose/issues/1823

ibuildthecloud:
According to conversations I had with @shykes the first priority of this project is strict conformance to the behaviour of the main docker-compose.  For this reason this is a valid change.  The incompatibility with volumes will be addressed in docker/compose#1823 and then ilbcompose will change as docker-compose does too.

@mssola Would you mind adding an integration test (https://github.com/docker/libcompose/blob/master/integration/basic_test.go)?

We need a bigger change than just this though.  Libcompose is designed to be ran in many contexts (embedded in a CLI, a GUI, a server side application, etc).  Especially for server side we need to give consideration to things that interact with the host.  There is already an interface call ConfigLookup that was designed to abstract away the logic of the looking up config files for extends.  Maybe we should look at enhancing this interface to resolve paths too.  Maybe instead of ConfigLookup we rename it to ResourceLookup and add a method for ResolvePath().  We can then just always unconditionally pass the first path of foo:bar to ResolvePath() and use the result of that.  @mssola WDTY?

It also might be easier to do this logic in docker.Container.createContainer() before we call ConvertToApi.  ConvertToApi() I think should be more straight translation of values, and does not interpret them.  The interpretation may be better done in the docker.Container struct.

mssola:
@ibuildthecloud I've added the integration tests already :)

Now, I'm afraid I don't quite understand your thought there. Even if it's used in many different contexts, the configuration still has to be picked from a config file, and we would still want to convert relative paths to absolute paths anyways. So I don't understand right now why do we need such abstraction :/

Moreover, let's say that we want to implement said abstraction. If you want to call this new `ResolvePath` method unconditionally before converting to the API, it means that on `createContainer` we should do something quite verbose such as:
```go
c.service.context.ResourceLookup.ResolvePath(c.service.serviceConfig.Volumes)
```
And, in turn, this function should do the exact same loop as in the `Convert` function, which feels like an overkill to me.

Another possibility is that we pass the docker.Service instead of the ServiceConfig to the `Convert` function. This way, in the loop of volumes inside of the `Convert` function we could call:
```go
vol := service.context.ResourceLookup.ResolvePath(v)
```
It's still a bit verbose for my taste, but I think that this would be a better solution.

ibuildthecloud:
@mssola Sorry I didn't notice the integration test, thanks!

The use case is basically that "." means the current directory or relative to something.  Today that is the filesystem of the client.  Imagine I post the compose template to a service (daemon), what is "current directory" now.  How can the service resolve ".", it shouldn't be a call to `Abs(...)`.  This is why Docker itself doesn't support relative paths because they aren't portable.  We need a way to address this in libcompose.

I understand the verbosity in the implementation.  What if we just added a second param to Convert() for a `ResourceLookup`.  If the lookup is nil we could probably just not resolve paths.  Then Convert can call the ResourceLookup to do path resolving.

mssola:
@ibuildthecloud now I get it, thanks. I've updated the patch accordingly ;)

ibuildthecloud:
LGTM

ibuildthecloud:
@mssola It just dawned on me, is ./ suppose to be relative to the current path or the docker-compose.yml file?

thaJeztah:
In compose, volume paths are relative to the docker-compose.yml; see https://github.com/docker/compose/issues/1823#issuecomment-129412364

mssola:
So, what about this ? :)

gdevillele:
@aanand will relative volume paths stay relative to the compose.yml file in the compose project?
Just asking for confirmation so we know what to do, since libcompose has to have the same behavior as compose. Thx.

dnephin:
Yes, paths in the compose.yml should always be relative to the file

ibuildthecloud:
@mssola The code needs to be updated to be relative to the project file.  The other method on ResourceLookup, Lookup() has the logic to do [relative lookups](https://github.com/docker/libcompose/pull/8/files#diff-45003f5c4683004dae98b286961ff44aR33).  I would just reuse that logic somehow as opposed to do `Abs()`.  Thanks for your work and patience.

mssola:
@ibuildthecloud I'm updating the patch and now I'm using the `ComposeFile` attribute to make it relative to the compose file. However, as I realized in the integration tests, this attribute might be set to "-", meaning that the compose file is loaded from the stdin. Am I correct? In this context, what is "relative to the project file" path ? :D

thaJeztah:
> However, as I realized in the integration tests, this attribute might be set to "-", meaning that the compose file is loaded from the stdin. Am I correct? In this context, what is "relative to the project file" path ? :D

Good question, never tried actually how compose handles that

@dnephin do you know?

dnephin:
`docker-compose` will use the current working directory if the file is specified as stdin (`-`)

https://github.com/docker/compose/blob/1c823597bb6c64de9cc320c54f6aa5ca1feff68f/compose/config/config.py#L95

mssola:
@ibuildthecloud updated!

anibali:
@mssola when I tried your code I found that when relativePath ran (https://github.com/mssola/libcompose/blob/1172545cf10138ca32b831555c016a8e202bb28a/lookup/file.go#L19-L34) `file` was relative and hence the function returns a relative path. This meant that I was still unable to use relative paths for volumes unless I added `filepath.Abs(file)`.

mssola:
I've updated the patch with what @anibali said, and I've also added a warning when relative paths don't start with "./", as implemented in the PR docker/compose#1833

mssola:
@vdemeester @ibuildthecloud could you review this ? :)

mssola:
I've rebased the code to current master.

ibuildthecloud:
@mssola I feel quite bad that this PR has taken so long to merge.  Can you rebase again?

mssola:
@ibuildthecloud I've been quite busy lately, I'll rebase this asap ;)

vdemeester:
@mssola carrying in #138, thanks a lot for your work ! (it's gonna be your commit in :wink:)

PR #9, opened 2015-08-06
Rename the binaries to libcompose-cli

dave-tucker:
Signed-off-by: Dave Tucker <dt@docker.com>

### DIFF
--- a/cli/main/main.go
+++ b/cli/main/main.go
@@ -14,8 +14,8 @@ func main() {
 	factory := &dockerApp.ProjectFactory{}
 
 	app := cli.NewApp()
-	app.Name = "docker-compose"
-	app.Usage = "Fast, isolated environments using Docker."
+	app.Name = "libcompose-cli"
+	app.Usage = "Command line interface for libcompose."
 	app.Version = version.VERSION + " (" + version.GITCOMMIT + ")"
 	app.Author = "Docker Compose Contributors"
 	app.Email = "https://github.com/docker/libcompose"
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 /main/main
 /docker-compose*
+/libcompose-cli*
 *.log
 *.swp
 bundles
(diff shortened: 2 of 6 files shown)


cli/main/main.go
```diff
@@ -14,7 +14,7 @@ func main() {
 	factory := &dockerApp.ProjectFactory{}
 
 	app := cli.NewApp()
-	app.Name = "docker-compose"
+	app.Name = "libcompose-cli"
 	app.Usage = "Fast, isolated environments using Docker."
```
aanand: Maybe this description should also change to something like "Command-line interface to libcompose"

aanand:
LGTM

ibuildthecloud:
Why would we not want this to be called `docker-compose`.  I think the intention is that if possible this could eventually become the default implementation of compose.  I'd be more in favor of calling this `docker-compose-experimental`.

bfirsh:
I have no opinions on the name, but I think the intention behind this is that the libcompose version of docker-compose is not _quite_ compatible with the Python CLI, so tools that assume there's something called docker-compose that behaves in a certain way will break.

@dave-tucker @aanand Is that right?

ibuildthecloud:
@bfirsh You make an excellent point.  I agree it shouldn't be called `docker-compose` as that is a direct conflict.  I think I still prefer something like `docker-compose-experimental` over `libcompose-cli` because it more clearly states to the user what it is.  My concern with libcompose-cli is that people might look at libcompose as some completely new fancy thing and not understand it at all.  If we are experimenting, we need user to try stuff :)

ibuildthecloud:
At the end of the day the name isn't important and we all agree on that it should be renamed.  So I'm cool with this PR but the tests will break now.

https://github.com/docker/libcompose/blob/master/integration/common.go#L35

I'm really hesitant to merge anything until we get CI running.  So maybe we just hold off until I can get that running?  @jfrazelle Halp!  How do we get this hooked up to Jenkins (I can ping you on IRC monday)

jfrazelle:
Just need GordonTheTurtle to have access to the repo, and 2 web hooks I can get you

imikushin:
Maybe just `compose`?

ibuildthecloud:
@dave-tucker Can you rebase?

dave-tucker:
@ibuildthecloud done. integration tests should now pass. also addressed @aanand's feedback. if needed, i'll happily change the name from `libcompose-cli` to whatever - i have no preference.

ibuildthecloud:
LGTM

PR #16, opened 2015-08-13
Support rebuilding containers

ibuildthecloud:
If you set Context.Rebuild=true then containers will be rebuilt if their
hash is out of sync

### DIFF
--- a/cli/command/command.go
+++ b/cli/command/command.go
@@ -65,6 +65,10 @@ func UpCommand(factory app.ProjectFactory) cli.Command {
 				Name:  "d",
 				Usage: "Do not block and log",
 			},
+			cli.BoolFlag{
+				Name:  "rebuild",
+				Usage: "Rebuild containers",
+			},
 		},
 	}
 }
@@ -217,6 +221,7 @@ func Populate(context *project.Context, c *cli.Context) {
 		context.Log = true
 	} else if c.Command.Name == "up" {
 		context.Log = !c.Bool("d")
+		context.Rebuild = c.Bool("rebuild")
 	} else if c.Command.Name == "stop" || c.Command.Name == "restart" || c.Command.Name == "scale" {
 		context.Timeout = c.Int("timeout")
 	} else if c.Command.Name == "kill" {
--- a/docker/container.go
+++ b/docker/container.go
@@ -95,6 +95,48 @@ func name(names []string) string {
 	return current[1:]
 }
 
+func (c *Container) Rebuild(imageName string) (*dockerclient.Container, error) {
+	info, err := c.findInfo()
+	if err != nil {
+		return nil, err
+	} else if info == nil {
+		return nil, fmt.Errorf("Can not find container to rebuild for service: %s", c.service.Name())
+	}
+
+	hash := info.Config.Labels[HASH.Str()]
+	if hash == "" {
+		return nil, fmt.Errorf("Failed to find hash on old container: %s", info.Name)
+	}
+
+	name := info.Name[1:]
+	new_name := fmt.Sprintf("%s_%s", name, info.Id[:12])
+	deleted := false
+	logrus.Debugf("Renaming %s => %s", name, new_name)
+	if err := c.client.RenameContainer(name, new_name); err != nil {
+		logrus.Errorf("Rename failed, deleting %s", name)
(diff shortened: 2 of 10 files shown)


ibuildthecloud:
I've found some bugs in this implementation.  I'll be pushing fixes in a bit.

ibuildthecloud:
This adds support for rebuilding containers on `up`.  This is a pretty large change.  It tried to add as many tests as I could to cover the corner cases.  Let me know if you think of any more.

This change makes it suck that when we create a container we hash the config and save the hash on the container as a label.  If the config changes we know we need to rebuild the container.  We also look to see if the `Config.Image` has changed to point at a different image id since it the container was created.

@aduermael @gdevillele

aduermael:
@ibuildthecloud is this aligned with that feature in py Compose 1.4.0? 
> By default, docker-compose up now only recreates containers for services whose configuration has changed since they were created. This should result in a dramatic speed-up for many applications.
(with this --force-recreate flag)

ibuildthecloud:
@aduermael This PR only rebuilds containers that have changed.  I didn't implement a `--force-recreate` flag.  I can do that as a separate PR.

I'm not totally sure on the nitty gritty details if it follows the same behavior.  I actually found it really really difficult to cover all the corner cases of dynamic rebuilding.  For example, if you do `volumes_from` and that container changed, do you rebuild?  Maybe @aanand can point me to the correct rebuilding behavior.  Currently if any configuration changes in the yml or if the image under the hood changes it will rebuild.  An example of image is if you do something like `docker-compose up`, then `docker pull`, then `docker-compose up` again.  If the `docker pull` pulled a new image that changed the tag, it should rebuild the container.

aduermael:
@ibuildthecloud  It would be better to do it in that same PR if possible no?
And yes, it would be great if you could point us to the correct rebuilding behavior @aanand. 
I guess these 2 flags should be supported: `--force-recreate` & `--no-recreate`

ibuildthecloud:
@aduermael I guess I could do both in this PR.  Right now the CLI params are wrong for this.  I'll update to be in line with `docker-compose`, so `--force-recreate` and `--no-recreate`.

aanand:
Sorry that the smart recreate logic isn't obvious - we should perhaps document it.

The basics:

- If anything in a service's configuration changes, it needs recreating.

- If the image layer pointed to by the service's image tag changes, it needs recreating.

- The above two are achieved by [creating a config dict](https://github.com/docker/compose/blob/1c73ff33fe12c1d93d29bd753f494fae9e2bb529/compose/service.py#L499-L506) for the service, serialising it to JSON (no whitespace, sorted keys), hashing it with SHA-256 and [storing it on the container as a label](https://github.com/docker/compose/blob/1c73ff33fe12c1d93d29bd753f494fae9e2bb529/compose/service.py#L957-L959) (`com.docker.compose.config-hash `).

    (The dict's exact format is not specified anywhere, but it probably should be if we want libcompose to generate the exact same hash.)

    The config hash is then [compared with those of the existing containers for the service](https://github.com/docker/compose/blob/1c73ff33fe12c1d93d29bd753f494fae9e2bb529/compose/service.py#L380-L387). If any container's config hash is different, the service has diverged and needs recreating. (That's overkill, of course - if some containers are up-to-date and others aren't, we could selectively recreate - but we kept it simple.)

- If any of a service's upstream dependencies (via `links` or `volumes_from` or `net: container`) need recreating, [it also needs recreating](https://github.com/docker/compose/blob/1c73ff33fe12c1d93d29bd753f494fae9e2bb529/compose/project.py#L294-L310).

ibuildthecloud:
@aduermael I have update this PR to reflect to support `--force-recreate` and `--no-recreate`

ibuildthecloud:
@imikushin I've ran into time issues with carry this work forward.  At this point just needs a rebase and the integration tests need to be broken up into the new format.  Do you want to take that on?

imikushin:
@ibuildthecloud Sure thing. Wanted to do just that after #78 got merged :wink:

ibuildthecloud:
Closing in favor of https://github.com/docker/libcompose/pull/89

thaJeztah:
Looks like you reopened?

vdemeester:
And I'll reclose :stuck_out_tongue_closed_eyes:. Github has reopened it with the commit I guess :wink:

PR #22, opened 2015-08-13
Fix event equality

ibuildthecloud:
Previously events we're based on strings so if the string version was the
same comparing two events would be true even though they were different
constants


gdevillele:
LGTM

PR #24, opened 2015-08-13
Fix validate-dco and temporarly remove validate-git-marks from default bundles

vdemeester:

### DIFF
--- a/script/.validate
+++ b/script/.validate
@@ -4,7 +4,7 @@ if [ -z "$VALIDATE_UPSTREAM" ]; then
 	# this is kind of an expensive check, so let's not do this twice if we
 	# are running more than one validate bundlescript
 	
-	VALIDATE_REPO='https://github.com/docker/machine.git'
+	VALIDATE_REPO='https://github.com/docker/libcompose.git'
 	VALIDATE_BRANCH='master'
 	
 	if [ "$TRAVIS" = 'true' -a "$TRAVIS_PULL_REQUEST" != 'false' ]; then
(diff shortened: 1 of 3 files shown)


ibuildthecloud:
Isn't this still useful if you run make locally?

vdemeester:
@ibuildthecloud I'm trying things out :sweat_smile:. But locally you would do `make validate` (or `make validate-dco`) and this would run `validate-dco` no problem. This is purely for the CI.

ibuildthecloud:
I merged https://github.com/docker/libcompose/pull/20, don't know if that impacts you much.

vdemeester:
@ibuildthecloud no no that won't :stuck_out_tongue_closed_eyes:. I'm gonna push in a few minutes something that should fix the build (on janky) :wink:.

Then we'll see how we handle `validate-dco` as it checks older commits and some are not signed-off so it fails :sweat:.

vdemeester:
@ibuildthecloud Fixed the problem without removing `validate-dco` which is very good. The problem was in `script/.validate`. 

There is a weird stuff with `validate-git-marks` though. I'll try one more thing to see if it builds. If not, I think we should temporarly remove it from the default bundle (so the build is green), and I'll fix that tomorrow morning (Europe/Paris timezone). WDYT ?

ibuildthecloud:
I'm fine with disabling git marks temporarily.  The script is more useful when you have lots of docs.  For go files they obviously won't compile so you don't need the validation.  Being that we have no docs :) we are safe atm.

PR #27, opened 2015-08-14
Add unit-tests (and thus coverage) to lookup pkg

vdemeester:
Add unit coverage to the package `lookup`, to start small 🐧.

```bash
+ go test -cover -coverprofile=cover.out ./lookup
ok      github.com/docker/libcompose/lookup     0.005s  coverage: 100.0% of statements
```

🐸

### DIFF
--- a/lookup/file_test.go
+++ b/lookup/file_test.go
@@ -1,0 +1,65 @@
+package lookup
+
+import (
+	"io/ioutil"
+	"path/filepath"
+	"testing"
+)
+
+type input struct {
+	file       string
+	relativeTo string
+}
+
+func TestLookupError(t *testing.T) {
+	invalids := map[input]string{
+		input{"", ""}:                             "read .: is a directory",
+		input{"", "/tmp/"}:                        "read /tmp: is a directory",
+		input{"file", "/does/not/exists/"}:        "open /does/not/exists/file: no such file or directory",
+		input{"file", "/does/not/something"}:      "open /does/not/file: no such file or directory",
+		input{"file", "/does/not/exists/another"}: "open /does/not/exists/file: no such file or directory",
+		input{"/does/not/exists/file", "/tmp/"}:   "open /does/not/exists/file: no such file or directory",
+		input{"does/not/exists/file", "/tmp/"}:    "open /tmp/does/not/exists/file: no such file or directory",
+	}
+
+	fileConfigLookup := FileConfigLookup{}
+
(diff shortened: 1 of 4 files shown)


lookup/file_test.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	if err = ioutil.WriteFile(tmpFile2, []byte("content2"), 0755); err != nil {
+		t.Fatal(err)
+	}
+
+	fileConfigLookup := FileConfigLookup{}
+
+	valids := map[input]string{
+		input{"file1", tmpFolder + "/"}:     "content1",
+		input{"file2", tmpFolder + "/"}:     "content2",
```
vdemeester: This felt weird to write :sweat_smile:.

lookup/simple_env_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"testing"
+)
+
+func TestOsEnvLookup(t *testing.T) {
+	osEnvLookup := &OsEnvLookup{}
+
+	envs := osEnvLookup.Lookup("PATH")
+	if len(envs) != 1 {
```
vdemeester: Probably should also test with a setup environment variables a test the content though.

lookup/simple_env.go
```diff
@@ -3,14 +3,12 @@ package lookup
...
 	"os"
-
-	"github.com/docker/libcompose/project"
 )
 
 type OsEnvLookup struct {
 }
 
-func (o *OsEnvLookup) Lookup(key, serviceName string, config *project.ServiceConfig) []string {
+func (o *OsEnvLookup) Lookup(key string) []string {
```
vdemeester: Simplified the method signature as serviceName and project.ServiceConfig are not needed at all there. I *reported* that to `EnvironmentLookup` too.

ibuildthecloud: @vdemeester I've embedded libcompose in other projects and the other values are quite useful.  I know the default impl doesn't use it but it is nice if you want to do context based lookup of env vars.  Also, just a side note, `ServiceConfig` doesn't actually include the service name in it so a lot of times we pass name+config to give full context.

vdemeester: Ok, sounds fair enough :wink: I'll put them back (and add a little godoc on it :wink:)

ibuildthecloud:
LGTM

gdevillele:
LGTM

PR #21, opened 2015-08-13
Setup validate-golint

vdemeester:
- Add a `validate-lint` target (and `script/validate-lint` that does the
  job)
- Add the first package that is covered by it : `version`.

For the demo :

```bash
λ make validate-lint  
# […]
Successfully built 875de57690ce
docker run --rm -it --privileged -e OS_PLATFORM_ARG -e OS_ARCH_ARG -e DOCKER_TEST_HOST -e TESTFLAGS -v "/home/vincent/src/docker/libcompose/bundles:/go/src/github.com/docker/libcompose/bundles" "libcompose-dev:14-setup-validate-golint" ./script/make.sh validate-lint
---> Making bundle: validate-lint (in .)
Congratulations!  All Go source files have been linted.
```

Related to #14 🐸

PR #31, opened 2015-08-17
Linting `cli` package

vdemeester:
Updated packages: `cli/app`, `cli/command`, `cli/docker/app`, `cli/logger`, `cli/main`.

Wating for #21 to be merged but that opens it to reviews already :wink:.

🐸

### DIFF
--- a/cli/docker/app/factory.go
+++ b/cli/docker/app/factory.go
@@ -8,9 +8,11 @@ import (
 	"github.com/docker/libcompose/project"
 )
 
+// ProjectFactory is a struct that hold the app.ProjectFactory implementation.
 type ProjectFactory struct {
 }
 
+// Create implements ProjectFactory.Create using docker client.
 func (p *ProjectFactory) Create(c *cli.Context) (*project.Project, error) {
 	context := &docker.Context{}
 	context.LoggerFactory = logger.NewColorLoggerFactory()
--- a/cli/logger/color_logger.go
+++ b/cli/logger/color_logger.go
@@ -9,23 +9,27 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 )
 
+// ColorLoggerFactory implements logger.Factory interface using ColorLogger.
 type ColorLoggerFactory struct {
 	maxLength int
 	tty       bool
 }
 
+// ColorLogger implements logger.Logger interface with color support.
 type ColorLogger struct {
 	name        string
 	colorPrefix string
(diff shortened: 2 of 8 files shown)


cli/docker/app/factory.go
```diff
@@ -8,10 +8,12 @@ import (
 	"github.com/docker/libcompose/project"
 )
 
-type ProjectFactory struct {
+// DockerProjectFactory is a struct that hold the app.ProjectFactory implementation.
+type DockerProjectFactory struct {
```
vdemeester: I renamed that as it felt a little weird to have the exact same name for the interface and the struct *implementing* it. If it breaks outside uses of the lib, tell me, I'll revert that change.

cli/logger/color_logger.go
```diff
@@ -9,23 +9,27 @@ import (
...
 )
 
+// ColorLoggerFactory implements logger.Factory interface using ColorLogger.
 type ColorLoggerFactory struct {
 	maxLength int
 	tty       bool
 }
 
+// ColorLogger implements logger.Logger interface with color support.
 type ColorLogger struct {
```
vdemeester: I think this one could be exported, if not used by a project depending on `libcompose` — it then depends on wether we want it to be exported or not :wink:.

cli/docker/app/factory.go
ibuildthecloud: It's pretty common in go to do this because any time you reference the type you also include package.  So calling code would look like `dockerApp.ProjectFactory`.   Making the type have the package name in it then makes it redundant `dockerApp.DockerProjectFactory`.

cli/logger/color_logger.go
ibuildthecloud: I think I'd prefer it to stay exported.

cli/docker/app/factory.go
vdemeester: Oh.. I didn't see it like that. I'll revert this one, agre with you, it makes it redundant.

cli/logger/color_logger.go
vdemeester: ok :-)

vdemeester:
@ibuildthecloud Updated according to your comments :wink:.

gdevillele:
LGTM :+1:

aduermael:
LGTM

vdemeester:
:tada: :wink:

PR #39, opened 2015-08-24
Implement container_name support

### LINKED ISSUE #28: container_name is not supported
`container_name` was added to docker-compose 1.4 is not supported in libcompose

vdemeester:
Fixes #28 — This implements `container_name` support as in docker-compose 1.4.

> Went with option 1 - a service with a custom name simply can't be scaled beyond 1 container. We let the Docker daemon enforce that.

This does the same as `docker-compose` : if `container_name` is present and we do a `scale` on it should fail but this will come from the `docker` side. This means one container will start, and as libcompose does not (yet ?) stop running containers in case of failure, it would stays up.

There is room for improvement in the integration tests, but that probably will be for another PR.

### DIFF
--- a/docker/service.go
+++ b/docker/service.go
@@ -86,7 +86,17 @@ func (s *Service) constructContainers(create bool, count int) ([]*Container, err
 
 	client := s.context.ClientFactory.Create(s)
 
-	namer := NewNamer(client, s.context.Project.Name, s.name)
+	var namer Namer
+
+	if s.serviceConfig.ContainerName != "" {
+		if count > 1 {
+			logrus.Warnf(`The "%s" service is using the custom container name "%s". Docker requires each container to have a unique name. Remove the custom name to scale the service.`, s.name, s.serviceConfig.ContainerName)
+		}
+		namer = NewSingleNamer(s.serviceConfig.ContainerName)
+	} else {
+		namer = NewNamer(client, s.context.Project.Name, s.name)
+	}
+
 	defer namer.Close()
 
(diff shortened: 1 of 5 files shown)


docker/service.go
```diff
@@ -103,12 +113,11 @@ func (s *Service) constructContainers(create bool, count int) ([]*Container, err
...
 			if err != nil {
 				return nil, err
-			} else {
-				logrus.Debugf("Created container %s: %v", dockerContainer.Id, dockerContainer.Names)
 			}
+			logrus.Debugf("Created container %s: %v", dockerContainer.Id, dockerContainer.Names)
 		}
 
-		result = append(result, NewContainer(client, containerName, s))
+		result = append(result, c)
```
vdemeester: I felt like this was a duplicate from https://github.com/docker/libcompose/pull/39/files#diff-9369bcb8195200a9e0e57e4fe7dda0b2L95 : we just call `NewContainer` with the same parameter, so :stuck_out_tongue:.

aduermael:
Hey @vdemeester, I reviewed this one, it's almost good to be merged, could you just resolve conflicts?

vdemeester:
@aduermael done :wink: (Github should add a "not mergeable" in the my Pull-Request view :sweat_smile:).

aduermael:
@vdemeester thanks! @gaetan is testing currently. :)

gdevillele:
LGTM 🐥

aduermael:
LGTM

Issue #32, opened 2015-08-17
libcompose should be go gettable/buildable/testable


aluzzardi:
As documented, the build and test process are currently driven by a `Makefile` performing actions in a container.

Since this is a library, it should be `go gettable/buildable/testable` out of the box without having to run a non-standard process.

There are apparently dependencies against non-pkg parts of docker which make it tricky to do so.

ibuildthecloud:
This custom makefile is just to use a DIND setup to build and test.  Besides the fact that this code is using non-pkg docker, it should be compatible with go get.  The non-pkg code needs to be resolved by pulling that code out of docker and into a separate project.  Which I believe is being worked on for 1.9.

aluzzardi:
I think the main offender is the `github.com/docker/docker/registry` import in `docker/container.go`.

That's preventing me to build a "hello world" using `libcompose`:

```
$ go build
../../docker/docker/registry/registry.go:21:2: cannot find package "github.com/docker/docker/autogen/dockerversion" in any of:
	/usr/local/go/src/github.com/docker/docker/autogen/dockerversion (from $GOROOT)
	/root/go/src/github.com/docker/docker/autogen/dockerversion (from $GOPATH)
```

Other than that, this is pretty awesome!

ibuildthecloud:
This is related to https://github.com/docker/docker/issues/11699.  I don't know if that issue is going to ever be resolved.  I'd guess the official solution would be "don't import that package."  So this goes back to the fact this code is uses non-pkg code from Docker.  The offending packages are:

```
github.com/docker/docker/cliconfig
github.com/docker/docker/opts
github.com/docker/docker/api
github.com/docker/docker/utils
github.com/docker/docker/graph/tags
github.com/docker/docker/runconfig
```

We can probably go through and start removing the dependency on these but the majority of the reason that the code is imported is because this library has to duplicate the same logic.  @icecrime This is in part why I'm really interested in working on the server/client API project.  A lot of "util" code should move there also.

vdemeester:
@ibuildthecloud The more *problematic* package on the list are `runconfig` and `registry. The server/client API project should definitely help resolving this issue, but it will depend what will be there.

On the long term the server/client API should help a lot to remove duplication. On the short term, there is a few package dependencies that we can remove (by duplicating the code a bit).

ibuildthecloud:
I hate duplicating code.  Maybe an incremental step would be to start putting in PRs to docker engine to move more common functionality to pkg.

[...]

PR #43, opened 2015-08-31
Add Installation step for Libcompose Developers

bohai:
Add Installation step for Libcompose Developers
Signed-off-by: Bo Hai <bohai@huawei.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ func main() {
 
 ## Building
 
+### Create binary for all platform
 You need Docker and ``make`` and then run the ``binary`` target. This
 will create binary for all platform in the `bundles` folder. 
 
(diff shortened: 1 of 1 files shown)


bohai:
Related issue: https://github.com/docker/libcompose/issues/32

bohai:
ping @aanand @aduermael @ibuildthecloud  @gdevillele

aanand:
Not sure I see the point of this versus just running `make binary`. (Though I find it strange that that builds *all* binaries by default.)

vdemeester:
As @aanand said, not sure this is that related with #32 issue : the issue basically says "we can't do go get github.com/docker/libcompose and have something that work as there is dependencies to the non *pkg* packages that depends on the dockerversion generated code".

bohai:
@aanand vdemeester

I think godep is not necessary.
In fact we only need to setup godep path to GOPATH.
Then we can use "go build" command to get the cli binary.

vdemeester:
@bohai true, but https://github.com/docker/libcompose/issues/32#issuecomment-131998164 is all about the fact we currently can't use it because it depends on stuff that depends on `docker/docker/autogen` package.

But you're right, we might even need godep (especially in the Dockerfile).

gdevillele:
so what do we do with this PR?
Do some of you think we should wait for [#32](https://github.com/docker/libcompose/issues/32) to be solved?

vdemeester:
@gdevillele I think it's nice to have something about "How to build libcompose without docker", and this PR would be the answer to that (with little changes). I think #32 will take quite some time to be fixed so we definitly should not wait :wink:.

So either we want something about "How to build libcompose without docker" and we could merge this, either we don't and we should close it. :sweat_smile:

dnephin:
I'm -1 on this. I think if we're going to document a workflow it should be the make+docker workflow, not one that runs on the host.

vdemeester:
@bohai thank you for the PR, but I'm gonna close this one for now. When #32 is *handled* it's not gonna needed anymore, and if we move away from `godep` (which is gonna be likely the case)

Feel free to comment if you think it should be re-open and re-discussed.

PR #44, opened 2015-09-02
make use of Dockerfile option

flower-pot:
Only a tiny addition, but since the information is already there, why not make use of it.


aduermael:
Thanks @flower-pot!
LGTM

gdevillele:
LGTM 🐭

PR #45, opened 2015-09-02
add run command

flower-pot:
Mimicking the docker-compose run command in form of

    $ libcompose <service> <command>

Signed-off-by: Frederic Branczyk <fbranczyk@gmail.com>

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -160,6 +160,30 @@ func ProjectPull(p *project.Project, c *cli.Context) {
 	}
 }
 
+// ProjectRun runs a given command within a service's container.
+func ProjectRun(p *project.Project, c *cli.Context) {
+	if len(c.Args()) == 1 {
+		logrus.Fatal("No service specified")
+	}
+	if len(c.Args()) <= 1 {
+		logrus.Fatal("No command specified")
+	}
+
+	serviceName := c.Args()[0]
+	commandParts := c.Args()[1:len(c.Args())]
+
+	if _, ok := p.Configs[serviceName]; !ok {
+		logrus.Fatalf("%s is not defined in the template", serviceName)
+	}
+
+	exitCode, err := p.Run(serviceName, commandParts)
+	if err != nil {
+		logrus.Fatal(err)
+	}
+
+	os.Exit(exitCode)
+}
+
 // ProjectDelete delete services.
 func ProjectDelete(p *project.Project, c *cli.Context) {
 	if !c.Bool("force") && len(c.Args()) == 0 {
--- a/cli/command/command.go
+++ b/cli/command/command.go
@@ -158,6 +158,16 @@ func ScaleCommand(factory app.ProjectFactory) cli.Command {
 	}
 }
 
+// RunCommand defines the libcompose run subcommand.
+func RunCommand(factory app.ProjectFactory) cli.Command {
+	return cli.Command{
+		Name:   "run",
+		Usage:  "Run a one-off command",
+		Action: app.WithProject(factory, app.ProjectRun),
+		Flags:  []cli.Flag{},
+	}
+}
+
 // RmCommand defines the libcompose rm subcommand.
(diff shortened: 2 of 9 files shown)


aanand:
We should reuse the docker client's code for attaching to containers, rather than re-implementing it - it's a really complex process.

flower-pot:
I agree. Maybe I'll take a look at the docker client once I have some more time on my hand, for now we'll just keep it line by line.

flower-pot:
I think I'm going to try and replace the current wait mechanism with one that closes the whole execution once all goroutines of the log output are closed. From what I found this is done using a WaitGroup.

flower-pot:
So I just pushed a suggestion on how to solve the log synchronisation and showing only the relevant logs. Let me know what you think.

flower-pot:
I revised all my commits and squashed them. Now the only thing I would like to improve is to really attach the container instead of just printing the log output.

flower-pot:
@aanand Do you think it would make sense to essentially instantiate the docker cli and call the [attach command](https://github.com/docker/docker/blob/6733962a2ca66d3d4d58772c378981a9e7da0c15/api/client/attach.go#L19)?

aanand:
@flower-pot Yeah, essentially, though instantiating a CLI object is going to be ugly.

It might be worth trying to extract that code so it's more reusable and submitting it as a PR to docker/docker.

flower-pot:
I agree. I'm going to take a look at how we can do that.

ibuildthecloud:
@flower-pot I agree the CLI functionality should be pulled out.  It would nice if we could find out what the status of https://github.com/docker/docker/issues/15188 is.  Basically there was supposed to be an effort to refactor a lot of the CLI code.

I've been down this path before with other projects I've written and agree that instantiating the docker CLI is a nightmare.  I think the correct first steps would be to figure out how to refactor the code in docker/docker to put the attach logic we need into a reusable pkg.  It's just all the dirty TTY handling that we need.  It can probably be encapsulated pretty well.

flower-pot:
Do you think the core team is going to take care of that? It seems pretty time consuming and I'm probably not the right guy to tackle this as I'm not very familiar with the docker codebase.

flower-pot:
Alright ported to the new dockerclient and now using its attach functionality. What do you think?

vdemeester:
@flower-pot Sorry for the *late* review. Thanks for the PR, few consideration/comments :

- `docker-compose run` supports more flags than this but I think it's good to focus on the main behavior and then the additionnal falgs (I'll create an issue to keep track of diff between `docker-compose` and `libcompose`).
- By default `docker-compose run` start the container with `stdin` and a pseudo-terminal (something like `-ti` in the docker cli). This is not the case in this PR — when I do a `libcompose-cli run myservice /bin/bash`, I can't do anything in it.

I'll take a look at the code too :wink:

cpuguy83:
Here is swarm's hijack implementation: https://github.com/docker/swarm/blob/894d1abd639eedb8234caf63fbc458a6791563bb/api/utils.go#L133

It's a bit cleaner.

cpuguy83:
It's a little different because it's proxying, but probably useful.

flower-pot:
@vdemeester no problem, I haven't had time to revise it either. I think I had the problem that when opening stdin that the stream would never close therefore the execution never ended. But I have to take another look to make sure that's the case.

vdemeester:
@flower-pot still willing (or having time) to work on that ?

flower-pot:
I don't really have the time for it right now. If you don't want to use my changes you can close the PR.

vdemeester:
@flower-pot ok, I'm gonna carry it then :wink:, keeping your commit and iterate on it :+1: 

Thanks for the hard work so far, really appreciated :santa:

flower-pot:
@vdemeester Glad I could help :)

vdemeester:
Carrying in #133.

Issue #34, opened 2015-08-21
Strict yaml validation?


ahmetalpbalkan:
I have mistakes in my docker-compose.yml file such as mistyped keys (`fooooocommand`):

```yml
consul:
  fooooocommand: -server -node master2 -advertise 10.0.0.6 -join 10.0.0.4 -join 10.0.0.5
  image: progrium/consul
  ports:
  - 8300:8300
```

libcompose seems to be ignoring them, however docker-compose does strict validation and errors out. We probably need the same behavior here.

thaJeztah:
This PR was just merged in compose, and performs validation of the docker-compose.yml; https://github.com/docker/compose/pull/1808, could be useful when implementing the same in libcompose.

ahmetalpbalkan:
Yes, jsonschema is most certainly a legit solution here as well and it has some good golang libs too.

ibuildthecloud:
That's a great idea if we can share schema files.  I'll look for a Json schema validator in go, hopefully we don't have to write one.

ibuildthecloud:
But generally yes I agree with this issue.  We should be stricter because I mistype things myself all the time.

ibuildthecloud:
Okay, found this https://github.com/xeipuuv/gojsonschema. It's on github, it has a common name, so I think that means its production ready :)

thaJeztah:
> Okay, found this https://github.com/xeipuuv/gojsonschema. It's on github, it has a common name, so I think that means its production ready :)

It doesn't have tags or releases yet, so we may want to check that with the author of that project.

ahmetalpbalkan:
it'll probably just work :) why don't we give it a shot? anyone?

ibuildthecloud:
I'm all for it, worst case scenario is doesn't work.  We got nothing to loose :)

bfirsh:
See also: https://github.com/docker/compose/issues/2089

joshwget:
Working on a PR for this, I think [xeipuuv/gojsonschema](https://github.com/xeipuuv/gojsonschema) will be sufficient. I've ran into a few small bugs(xeipuuv/gojsonschema#72, xeipuuv/gojsonschema#73, xeipuuv/gojsonschema#74), but I think I can include some band-aids to get around them until they're fixed.

dnephin:
[...]

PR #47, opened 2015-09-07
Support interpolating environment variables

### LINKED ISSUE #40: Support interpolating environment variables
https://github.com/docker/compose/pull/1765

joshwget:
Fixes #40

### DIFF
--- a/integration/basic_test.go
+++ b/integration/basic_test.go
@@ -49,6 +49,66 @@ func (s *RunSuite) TestHelloWorld(c *C) {
 	c.Assert(cn.Name, Equals, "/"+name)
 }
 
+func (s *RunSuite) TestInterpolation(c *C) {
+	os.Setenv("IMAGE", "tianon/true")
+
+	p := s.CreateProjectFromText(c, `
+	test:
+	  image: $IMAGE
+	`)
+
+	name := fmt.Sprintf("%s_%s_1", p, "test")
+	testContainer := s.GetContainerByName(c, name)
+
+	p = s.CreateProjectFromText(c, `
+	reference:
+	  image: tianon/true
+	`)
+
+	name = fmt.Sprintf("%s_%s_1", p, "reference")
+	referenceContainer := s.GetContainerByName(c, name)
+
+	c.Assert(testContainer, NotNil)
+
+	c.Assert(referenceContainer.Image, Equals, testContainer.Image)
+
+	os.Unsetenv("IMAGE")
+}
+
+func (s *RunSuite) TestInterpolationWithExtends(c *C) {
+	os.Setenv("IMAGE", "tianon/true")
+	os.Setenv("TEST_PORT", "8000")
+
+	p := s.CreateProjectFromText(c, `
+	test:
+		extends:
+			file: ./assets/interpolation/docker-compose.yml
+			service: base
+		ports:
+			- ${TEST_PORT}
+	`)
+
+	name := fmt.Sprintf("%s_%s_1", p, "test")
+	testContainer := s.GetContainerByName(c, name)
+
+	p = s.CreateProjectFromText(c, `
+	reference:
+	  image: tianon/true
+		ports:
+		  - 8000
+	`)
+
+	name = fmt.Sprintf("%s_%s_1", p, "reference")
+	referenceContainer := s.GetContainerByName(c, name)
+
+	c.Assert(testContainer, NotNil)
+
+	c.Assert(referenceContainer.Image, Equals, testContainer.Image)
+
+	os.Unsetenv("TEST_PORT")
+	os.Unsetenv("IMAGE")
+}
+
 func (s *RunSuite) TestUp(c *C) {
 	p := s.ProjectFromText(c, "up", SimpleTemplate)
 
--- a/project/interpolation.go
+++ b/project/interpolation.go
@@ -1,0 +1,150 @@
+package project
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+
+	"github.com/Sirupsen/logrus"
+)
+
+func parseVariable(line string, pos int, mapping func(string) string) (string, int, bool) {
+	var buffer bytes.Buffer
+
+	for ; pos < len(line); pos++ {
+		c := line[pos]
+
+		switch {
+		case c == '_' || (c >= 'A' && c <= 'Z'):
+			buffer.WriteByte(c)
+		default:
+			return mapping(buffer.String()), pos - 1, true
+		}
+	}
+
+	return mapping(buffer.String()), pos, true
+}
+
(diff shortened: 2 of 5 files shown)


ibuildthecloud:
@joshwget You have one `go vet` issue

```
project/interpolation.go:66: unreachable code
```

Thanks for the contribution, I'll try to provide some feedback tonight.

project/interpolation_internal_test.go
```diff
@@ -0,0 +1,161 @@
+package project
```
dnephin: From what I've seen there is a convention to name the test file `interpolation_test.go`. What's the reason for deviating from that convention ?

joshwget: I thought it was necessary for internal tests. I'll rename it!

integration/basic_test.go
```diff
@@ -31,6 +31,32 @@ func (s *RunSuite) TestHelloWorld(c *C) {
 	c.Assert(cn.Name, Equals, "/"+name)
 }
 
+func (s *RunSuite) TestInterpolation(c *C) {
```
ibuildthecloud: Can you add another test or include it in this one that validates that interpolation happens for [extended templates](docs.docker.com/compose/yml/#extends) meaning that the parent and child templates both have variables in them.

project/interpolation.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	return nil
+}
+
+func Interpolate(config *rawServiceMap) error {
+	for k, v := range *config {
+		for k2, v2 := range v {
+			err := interpolate(k2, k, &v2, func(s string) string {
+				value := os.Getenv(s)
```
ibuildthecloud: We have an abstraction already in place for looking up env variables: https://github.com/docker/libcompose/blob/master/project/types.go#L203.  It is stored in the `project.Context`.  Please use that abstraction to actually resolve the env vars.  The reason for this is that if you are processing the template in a server side context you can not pull the environment variable values from the current environment, but somewhere else, like a saved environment.

The EnvironmentLookup interface does return multiple values, but I would always just take the first one if the result is greater than 0.

project/interpolation.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	return nil
+}
+
+func Interpolate(environmentLookup EnvironmentLookup, config *rawServiceMap) error {
+	for k, v := range *config {
+		for k2, v2 := range v {
+			err := interpolate(k2, k, &v2, func(s string) string {
+				values := environmentLookup.Lookup(s, k, nil)
```
joshwget: Wasn't sure what to do with the third argument here since a ServiceConfig isn't created until after interpolation is performed. This works fine using the OsEnvLookup implementation, which is currently the only one, but it might not be a good idea in general.

ibuildthecloud: Yeah, that makes sense.  I think it's fine to say that the serviceConfig could be nil and the impl of EnvironmentLookup so be programmed so.

ibuildthecloud:
LGTM

ping @aduermael @gdevillele

project/interpolation_test.go
```diff
@@ -0,0 +1,170 @@
...
+		"A": "ABC",
+		"X": "XYZ",
+		"E": "",
+	}
+
+	testInterpolatedLine(t, "ABC", "$A", variables)
+	testInterpolatedLine(t, "ABC", "${A}", variables)
+
+	testInterpolatedLine(t, "ABC DE", "$A DE", variables)
+	testInterpolatedLine(t, "ABCDE", "${A}DE", variables)
```
vdemeester: I would add a `testInterpolatedLine(t, "", "$ADE", variables)` case

project/interpolation_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	testInterpolatedLine(t, "$A", "$$A", variables)
+	testInterpolatedLine(t, "${A}", "$${A}", variables)
+
+	testInterpolatedLine(t, "$ABC", "$$${A}", variables)
+	testInterpolatedLine(t, "$ABC", "$$$A", variables)
+
+	testInterpolatedLine(t, "ABC XYZ", "$A $X", variables)
```
vdemeester: Same here, a `testInterpolatedLine(t, "ABCXYZ", "$A$X", variables)` case

vdemeester:
@joshwget looks good :wink:. Two things before merging :wink: :

- Could you rebase against `master`. You will need to `lint` your changes (should be small changes), you can check if it's valid using `make validate` (or just use `golint` and `go vet` :stuck_out_tongue_closed_eyes:).
- If I run the `interpolation/docker-compose.yml` with any environment defined, I get :
```bash
WARN[0000] The IMAGE variable is not set. Substituting a blank string. 
INFO[0000] [0/1] [base]: Starting
ERRO[0000] Failed Starting base : 500 Internal Server Error: No command specified
ERRO[0000] Failed to start: base : 500 Internal Server Error: No command specified
FATA[0000] 500 Internal Server Error: No command specified
```

Using `docker-compose` I get : 
```bash
# […] long stack >_<
    do_build=do_build,
  File "/code/compose/service.py", line 317, in ensure_image_exists
    self.image()
  File "/code/compose/service.py", line 332, in image
    return self.client.inspect_image(self.image_name)
  File "/code/.tox/py27/lib/python2.7/site-packages/docker/utils/decorators.py", line 18, in wrapped
    'image or container param is undefined'
docker.errors.NullResource: image or container param is undefined
```

The warning is good enough I think but I feel the `500` errors are a little bit weird.. But `docker-compose` doesn't do too much better it seems so.. for now I guess it's ok :sweat_smile:.

Anyway, this looks good :wink:.

vdemeester:
@joshwget what do you think of my comments on the tests ? :smile_cat: (I'm okay merging without it, i't just nits :wink:)

joshwget:
@vdemeester I think the 500 errors will go away when #34 is fixed, and I'm currently working on a PR for that! I also added the test cases you recommended.

vdemeester:
@joshwget ah you're right :stuck_out_tongue_closed_eyes: I might need some sleep or glasses :sweat_smile:.

LGTM

project/interpolation.go
```diff
@@ -0,0 +1,150 @@
...
+
+func parseInterpolationExpression(line string, pos int, mapping func(string) string) (string, int, bool) {
+	c := line[pos]
+
+	switch {
+	case c == '$':
+		return "$", pos, true
+	case c == '{':
+		return parseVariableWithBraces(line, pos+1, mapping)
+	case c >= 'A' && c <= 'Z':
```
dnephin: This is missing lowercase letters, which are valid variable names.

ibuildthecloud: I created https://github.com/docker/libcompose/pull/70 to address this.

PR #48, opened 2015-09-08
Warn if scaling a service that specifies a host port

joshwget:
Mimics the warning that Compose gives when trying to scale a service that specifies a port on the host.

### DIFF
--- a/docker/service.go
+++ b/docker/service.go
@@ -1,7 +1,9 @@
 package docker
 
 import (
+	"fmt"
 	"github.com/Sirupsen/logrus"
+	"github.com/docker/docker/pkg/nat"
 	"github.com/docker/libcompose/project"
 	"github.com/docker/libcompose/utils"
 )
@@ -251,6 +253,10 @@ func (s *Service) Log() error {
 // Scale implements Service.Scale. It creates or removes containers to have the specified number
 // of related container to the service to run.
 func (s *Service) Scale(scale int) error {
+	if s.specificiesHostPort() {
+		logrus.Warnf("The \"%s\" service specifies a port on the host. If multiple containers for this service are created on a single host, the port will clash.", s.Name())
+	}
+
 	foundCount := 0
 	err := s.eachContainer(func(c *Container) error {
 		foundCount++
@@ -304,4 +310,22 @@ func (s *Service) Containers() ([]project.Container, error) {
 	}
 
 	return result, nil
+}
+
+func (s *Service) specificiesHostPort() bool {
+	_, bindings, err := nat.ParsePortSpecs(s.Config().Ports)
+
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	for _, portBindings := range bindings {
+		for _, portBinding := range portBindings {
+			if portBinding.HostPort != "" {
+				return true
+			}
+		}
+	}
+
+	return false
 }
--- a/docker/service_test.go
+++ b/docker/service_test.go
@@ -1,0 +1,27 @@
+package docker
+
+import (
+	"github.com/docker/libcompose/project"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestSpecifiesHostPort(t *testing.T) {
+	servicesWithHostPort := []Service{
+		{serviceConfig: &project.ServiceConfig{Ports: []string{"8000:8000"}}},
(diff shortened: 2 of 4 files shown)


vdemeester:
Hi @joshwget, thanks for the contribution :) — would be cool to add some integration test.

project/types.go
```diff
@@ -237,6 +237,7 @@ type Service interface {
 	DependentServices() []ServiceRelationship
 	Containers() ([]Container, error)
 	Scale(count int) error
+	SpecificiesHostPort() bool
```
dnephin: Should this really be part of the public interface? It seems like more of an internal helper function.

docker/service.go
```diff
@@ -273,3 +279,13 @@ func (s *Service) Containers() ([]project.Container, error) {
 
 	return result, nil
 }
+
+func (s *Service) SpecificiesHostPort() bool {
+	for _, port := range s.serviceConfig.Ports {
+		if strings.ContainsRune(port, ':') {
+			return true
+		}
+	}
```
dnephin: I think we need to fix this check in `docker-compose` because I don't think this is accurate. If ports is `127.0.0.1::3600` this will return true, but there is no host port specified.

joshwget: Thanks for catching this! It looks like <IP>::<port> is the only case that fails, so it should be sufficient to check that "::" also isn't in the port string, right?

project/types.go
joshwget: I don't feel too strongly about it either way. It just seemed like something that might be useful.

joshwget:
Hi @vdemeester, no problem! Good idea, I'll work on those.

project/types.go
ibuildthecloud: I would definitely removed this from the interface.  This is a property of the Docker implementation and no the definition of a Service itself.

docker/service.go
ibuildthecloud: Refer to https://github.com/docker/libcompose/blob/master/docker/convert.go#L60 for the proper way to parse a port spec.

vdemeester:
Overalls looks good :wink:. I'm wondering about the change in the `integration/common.go` but it's beyond the scope of this PR.

ping @aduermael @gdevillele

vdemeester:
@joshwget Could you squash your commit ?
Other than that, LGTM :+1:

docker/service.go
```diff
@@ -273,3 +279,21 @@ func (s *Service) Containers() ([]project.Container, error) {
 
 	return result, nil
 }
+
+func (s *Service) SpecificiesHostPort() bool {
```
ibuildthecloud: Can this be moved to private, not a public method.

ibuildthecloud:
@joshwget Can you squash and rebase?  I think this may have been branched before @vdemeester added a lot of the vet/lint stuff.

vdemeester:
Ohh true, same as the other PR :sweat_smile: :wink:.

joshwget:
@ibuildthecloud @vdemeester I'm getting a lot of lint errors in integration/common.go that aren't relevant to the changes I made. Should I fix those in this PR or a separate one?

ibuildthecloud:
@joshwget That doesn't totally makes sense.  Are you sure the rebase work right?  Does the git history and changes look right? Master is clean if you run `make validate`

joshwget:
@ibuildthecloud `integration/common.go` from master doesn't pass golint. This only seems to cause `make validate` to fail when a change to the file is made and then committed, which is the case for this PR.

vdemeester:
ah @joshwget righ, integration is not linted ; it's the integration test code and not actually libcompose code. The `script/validate-*` explicitely skip this one :sweat_smile:.

joshwget:
@vdemeester I might just be doing something wrong, but it seems that `make validate` is actually calling `golint` on `integration/common.go`, causing it to fail. I didn't have the same problem when I rebased my other two PRs, so this only seems to happen when a commit includes a change to `integration/common.go`.

vdemeester:
@joshwget strange, it should : https://github.com/docker/libcompose/blob/master/script/validate-lint#L9

vdemeester:
@joshwget could you rebase against master ? It should be fixed now :wink:

ibuildthecloud:
LGTM, but needs rebase :(

vdemeester:
Yay, thanks again for your patience @joshwget :wink:

joshwget:
No problem! Thanks @vdemeester and @ibuildthecloud

PR #62, opened 2015-09-29
Call validate-lint on changed files

### LINKED ISSUE #14: Proposal : Enable `golint` on the code base
Taking the description directly from docker#14756.

> We want to enable golint on our codebase for several reasons:
> 
> - We want to improve code quality
> - We need objective filters on quality to help us discriminate bad pull requests

There is some work to do, but I think it's would be better to do it as *early* as possible (easier to do). And as libcompose is intended to be used as a library more than just a cli, documentation and quality is import I think.

WDYT about that ?
If you think we should do it, we'll keep track of the work on this issue then :wink:.

Package (and linting status) :
- [X] cli/app #31
- [X] cli/command #31
- [X] cli/docker/app #31
- [X] cli/logger #31
- [X] cli/main
- [X] docker #51 
- [X] logger #50 
- [X] lookup
- [X] project #49 
- [X] utils #37
- [X] version

vdemeester:
Now that `golint` is done on all source files, let set up `golint` to check only changed files (like `validate-vet`).

Closes #14.

🐸

### DIFF
--- a/script/validate-lint
+++ b/script/validate-lint
@@ -1,29 +1,18 @@
 #!/bin/bash
 
+source "$(dirname "$BASH_SOURCE")/.validate"
+
 # We will eventually get to the point where packages should be the complete list
 # of subpackages, vendoring excluded, as given by:
 #
-# packages=( $(go list ./... 2> /dev/null | grep -vE "^github.com/docker/libcompose/Godeps" || true ) )
-
-packages=(
-    cli/app
-    cli/command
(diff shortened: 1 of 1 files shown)


dnephin:
LGTM

I'd like to use http://pre-commit.com/ to run some of these, but I'll create a separate issue for that.

ibuildthecloud:
LGTM

PR #64, opened 2015-10-06
Do not log fatal while parsing

ibuildthecloud:

### DIFF
--- a/project/merge.go
+++ b/project/merge.go
@@ -29,15 +29,12 @@ var (
 type rawService map[string]interface{}
 type rawServiceMap map[string]rawService
 
-// Merge read the specified byte array, which is the content of a yaml composefile,
-// and merge it into the current project.
-func Merge(p *Project, bytes []byte) (map[string]*ServiceConfig, error) {
+func mergeProject(p *Project, bytes []byte) (map[string]*ServiceConfig, error) {
 	configs := make(map[string]*ServiceConfig)
 
 	datas := make(rawServiceMap)
-	err := yaml.Unmarshal(bytes, &datas)
-	if err != nil {
-		logrus.Fatalf("Could not parse config for project %s : %v", p.Name, err)
+	if err := yaml.Unmarshal(bytes, &datas); err != nil {
+		return nil, err
 	}
 
 	for name, data := range datas {
@@ -50,7 +47,7 @@ func Merge(p *Project, bytes []byte) (map[string]*ServiceConfig, error) {
 		datas[name] = data
 	}
 
-	err = utils.Convert(datas, &configs)
+	err := utils.Convert(datas, &configs)
 	return configs, err
 }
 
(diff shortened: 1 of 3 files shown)


ibuildthecloud:
The causes badness if somebody uses libcompose in a server application and has a bad formatted file.  We would just `exit 1`, so yeah, that was bad.

dnephin:
A couple unit tests for this case would be great

ibuildthecloud:
@dnephin Updated and added a unit test.

project/merge.go
```diff
@@ -37,7 +37,8 @@ func Merge(p *Project, bytes []byte) (map[string]*ServiceConfig, error) {
 	datas := make(rawServiceMap)
 	err := yaml.Unmarshal(bytes, &datas)
 	if err != nil {
-		logrus.Fatalf("Could not parse config for project %s : %v", p.Name, err)
+		logrus.Errorf("Could not parse config for project %s : %v", p.Name, err)
+		return nil, err
```
dnephin: It seems like this `Merge` should be private?  Is it really part of the public interface? It's currently only called from one place in the same package, and from my reading seems like it shouldn't be public.

project/project.go
```diff
@@ -123,7 +123,8 @@ func (p *Project) Load(bytes []byte) error {
 	configs := make(map[string]*ServiceConfig)
 	configs, err := Merge(p, bytes)
 	if err != nil {
-		log.Fatalf("Could not parse config for project %s : %v", p.Name, err)
+		log.Errorf("Could not parse config for project %s : %v", p.Name, err)
+		return err
```
dnephin: Merge just logged this error, do we need to log it a second time?

project/project_test.go
```diff
@@ -19,3 +20,27 @@ func TestEventEquality(t *testing.T) {
...
+	}
+
+	p = NewProject(&Context{
+		ComposeBytes: []byte("not-garbage:\n  image: foo"),
+	})
+
+	err = p.Parse()
+	if err != nil {
+		t.Fatal(err)
+	}
```
dnephin: I always put the "success" case and "failure" case in separate tests so the output summary is more clear.  I'm not sure if this is a common practice in golang or not.

project/project_test.go
```diff
@@ -19,3 +20,27 @@ func TestEventEquality(t *testing.T) {
 		t.Fatal("Events match")
 	}
 }
+
+func TestBadFile(t *testing.T) {
```
dnephin: Could this be `TestParseWithBadFile()` so it's clear which function is being tested?

dnephin:
mostly looks good, just some really minor nits and questions

project/project.go
vdemeester: Agree on that, it seems this one is not needed :wink:

project/project_test.go
vdemeester: Same, I would split this test into `TestParseWithBadYaml` and `TestParseWithValidYaml`.

ibuildthecloud:
@dnephin @vdemeester Updated based on comment.  Everything should be addressed.

vdemeester:
LGTM 🐹

PR #65, opened 2015-10-07
Allow multiple calls to project methods without reinitializing project

ibuildthecloud:
Prior to this change if one was to do project.Create() and then
project.Start(), with the same project object, the second invocation
would do nothing.  This is because the initial list of services to
traverse was based off of p.reload.  This patch changes it such that
reload is not taken into account on start but only when we need to
restart the execution.

### DIFF
--- a/project/project.go
+++ b/project/project.go
@@ -136,16 +136,14 @@ func (p *Project) Load(bytes []byte) error {
 	return nil
 }
 
-func (p *Project) loadWrappers(wrappers map[string]*serviceWrapper) error {
-	for _, name := range p.reload {
+func (p *Project) loadWrappers(wrappers map[string]*serviceWrapper, servicesToConstruct []string) error {
+	for _, name := range servicesToConstruct {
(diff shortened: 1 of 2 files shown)


project/project.go
```diff
@@ -136,8 +136,8 @@ func (p *Project) Load(bytes []byte) error {
 	return nil
 }
 
-func (p *Project) loadWrappers(wrappers map[string]*serviceWrapper) error {
-	for _, name := range p.reload {
+func (p *Project) loadWrappers(wrappers map[string]*serviceWrapper, list []string) error {
+	for _, name := range list {
```
dnephin: What's in this list? could we get a better variable name or a comment about the contents?

ibuildthecloud: will do.  It is the list of service names to initialize.  FYI, I know `project.go` is highly unreadable.  I plan to rewrite it.

vdemeester:
LGTM 🐺

PR #66, opened 2015-10-07
Fix issue in which extending could end up with both build and image

ibuildthecloud:
According to Docker Compose docs

> In the case of build and image, using one in the local service causes
> Compose to discard the other, if it was defined in the original
> service.

The above was not being respected

### DIFF
--- a/project/merge_test.go
+++ b/project/merge_test.go
@@ -1,0 +1,148 @@
+package project
+
+import "testing"
+
+type NullLookup struct {
+}
+
+func (n *NullLookup) Lookup(file, relativeTo string) ([]byte, string, error) {
+	return nil, "", nil
+}
+
+func TestExtendsInheritImage(t *testing.T) {
+	p := NewProject(&Context{
+		ConfigLookup: &NullLookup{},
+	})
+
+	config, err := Merge(p, []byte(`
+parent:
+  image: foo
+child:
+  extends:
+    service: parent
+`))
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	parent := config["parent"]
+	child := config["child"]
+
+	if parent.Image != "foo" {
+		t.Fatal("Invalid image", parent.Image)
+	}
+
+	if child.Build != "" {
+		t.Fatal("Invalid build", child.Build)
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
ping @dnephin @vdemeester

project/merge_test.go
```diff
@@ -0,0 +1,148 @@
...
+		t.Fatal("Invalid image", parent.Image)
+	}
+
+	if child.Build != "" {
+		t.Fatal("Invalid build", child.Build)
+	}
+
+	if child.Image != "foo" {
+		t.Fatal("Invalid image", child.Image)
+	}
```
dnephin: It looks like we already have `testify.assert` as a dependency.

What do you think about making all of these use `assert.Equal()` ?

dnephin:
LGTM, but I think it would be nice to make use of testify.assert

project/merge_test.go
vdemeester: I'm not against `testify` if we already have it in our dependencies. @ibuildthecloud opinions ? :smiley_cat:

ibuildthecloud: I'd really, really, really like to use go-check.  I'm not a fan of testify and would like to move away from it.  I did start with testify for this code.  Are we cool with go-check?  Engine uses go-check

dnephin: I believe engine uses go-check for integration tests, and has a mix of testify.assert and regular checks in unit tests.

Anything is better than the default `t.Fatal()` stuff. I think that testify.assert doesn't require you to change the test suite, it's just functions you can use with the default suite.

ibuildthecloud: Yes, go-check is more intrusive.  But I can just use testify.assert.  I'll change that now.

vdemeester: Ok for go-check too :smile_cat:.

I can do a separate PR on migrating test to `go-check` then :wink:.

ibuildthecloud:
@dnephin Going forward I'll use testify.assert.  Rather not change something now and break it.

dnephin:
cool, works for me

PR #68, opened 2015-10-07
 A blank value for environment should be treated as unset

ibuildthecloud:
This fixes an issue in which the below would not lookup `A` in the
current environment.

```yaml
test:
  image: foo
  environment:
    A:
```

This PR is based on #65 because it had all the test stubs I needed.

### DIFF
--- a/project/project.go
+++ b/project/project.go
@@ -91,9 +91,12 @@ func (p *Project) CreateService(name string) (Service, error) {
 		parsedEnv := make([]string, 0, len(config.Environment.Slice()))
 
 		for _, env := range config.Environment.Slice() {
-			if strings.IndexRune(env, '=') != -1 {
+			parts := strings.SplitN(env, "=", 2)
+			if len(parts) > 1 && parts[1] != "" {
 				parsedEnv = append(parsedEnv, env)
 				continue
+			} else {
+				env = parts[0]
 			}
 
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
ping @dnephin @vdemeester

project/project.go
```diff
@@ -317,16 +318,25 @@ func (p *Project) startService(wrappers map[string]*serviceWrapper, history []st
 	return nil
 }
 
-func (p *Project) traverse(selected map[string]bool, wrappers map[string]*serviceWrapper, action wrapperAction, cycleAction serviceAction) error {
+func (p *Project) traverse(start bool, selected map[string]bool, wrappers map[string]*serviceWrapper, action wrapperAction, cycleAction serviceAction) error {
```
cpuguy83: boolean arguments are the devil.
This functional also takes a _lot_ of parameters.

ibuildthecloud: @cpuguy83 I agree somewhat.  I don't want to change this though right now.  The problem is that project.go really needs to be completely refactored.  It is impossible to read and not flexible enough.  Because of this I rather just address this immediate bug and then work on a refactor.

vdemeester:
@ibuildthecloud needs rebase :sweat:

vdemeester:
LGTM, carrying in #94.

PR #70, opened 2015-10-07
Support a-z and 0-9 in variable names

ibuildthecloud:

### DIFF
--- a/project/interpolation.go
+++ b/project/interpolation.go
@@ -8,6 +8,17 @@ import (
 	"github.com/Sirupsen/logrus"
 )
 
+func isNum(c uint8) bool {
+	return c >= '0' && c <= '9'
+}
+
+func validVariableNameChar(c uint8) bool {
+	return c == '_' ||
+		c >= 'A' && c <= 'Z' ||
+		c >= 'a' && c <= 'z' ||
+		isNum(c)
+}
+
 func parseVariable(line string, pos int, mapping func(string) string) (string, int, bool) {
 	var buffer bytes.Buffer
 
@@ -15,7 +26,7 @@ func parseVariable(line string, pos int, mapping func(string) string) (string, i
 		c := line[pos]
 
 		switch {
-		case c == '_' || (c >= 'A' && c <= 'Z'):
+		case validVariableNameChar(c):
 			buffer.WriteByte(c)
 		default:
 			return mapping(buffer.String()), pos - 1, true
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
@dnephin Can you confirm that the list of valid character is correct?  It is now a-z, A-Z, _ and 0-9.

dnephin:
`r'[_a-z][_a-z0-9]*'` is the pattern (used with `_re.IGNORECASE`), so I believe that is correct

project/interpolation_test.go
```diff
@@ -26,11 +27,20 @@ func testInvalidInterpolatedLine(t *testing.T, line string) {
...
+		"lower":       "WORKED",
+		"MiXeD":       "WORKED",
+		"split_VaLue": "WORKED",
+		"9aNumber":    "WORKED",
 	}
 
+	testInterpolatedLine(t, "WORKED", "$lower", variables)
+	testInterpolatedLine(t, "WORKED", "${MiXeD}", variables)
+	testInterpolatedLine(t, "WORKED", "${split_VaLue}", variables)
+	testInterpolatedLine(t, "WORKED", "$9aNumber", variables)
```
dnephin: In the python implementation this wouldn't work. I think the reason is that it's using the same logic as bash variables, which can't start with a number

ibuildthecloud: Okay, I can make thing starting with a number not work, but an embedded number would be fine?

ibuildthecloud:
@dnephin Updated to not allow starting with a number

thaJeztah:
Is this for env-vars? You may be interested in https://github.com/docker/docker/pull/16608 (and the linked issue)

dnephin:
@thaJeztah nope, not for env-vars, this is for variables in the compose file (it just happens we use very similar syntax).

project/interpolation.go
```diff
@@ -8,6 +8,17 @@ import (
 	"github.com/Sirupsen/logrus"
 )
 
+func isNum(c uint8) bool {
+	return c >= '0' && c <= '9'
+}
```
dnephin: It still kind of surprises me that there is no golang builtin for super common operations like this

dnephin:
LGTM

project/interpolation.go
ibuildthecloud: The joys of golang.

PR #73, opened 2015-10-07
Revert change in a75b53 to make ValidRemotes private

ibuildthecloud:
This constant was in use by code using libcompose

### DIFF
--- a/project/merge.go
+++ b/project/merge.go
@@ -13,7 +13,9 @@ import (
 )
 
 var (
-	validRemotes = []string{
+	// ValidRemotes list the of valid prefixes that can be sent to Docker as a build remote location
+	// This is public for consumers of libcompose to use
+	ValidRemotes = []string{
 		"git://",
 		"git@github.com:",
 		"github.com",
@@ -120,7 +122,7 @@ func resolveBuild(inFile string, serviceData rawService) (rawService, error) {
 		return serviceData, nil
 	}
 
-	for _, remote := range validRemotes {
+	for _, remote := range ValidRemotes {
 		if strings.HasPrefix(build, remote) {
 			return serviceData, nil
 		}


dnephin:
Could we expose a better interface than a list of constants? Why does the code need to know about this?

ibuildthecloud:
@dnephin Hmmm... The reason the code is here is that you can build from any valid Docker remote location, https://docs.docker.com/reference/api/docker_remote_api_v1.20/#build-image-from-a-dockerfile

Looking at the Docker compose docs doesn't seem like this is possible.  Is there a specific reason you aren't allowed to do a Docker build from a remote location?  By remote I mean the definition of the `remote` param on `POST /build`

dnephin:
No reason, there's an PR to add that support (that needs more work), but I don't know if anyone is actively working on it.

There's been very little demand for it, so it hasn't been a priority.

dnephin:
Sorry, my earlier question was, why does code external to libcompose need this constant

ibuildthecloud:
@dnephin The reason why is that you can supply a custom Builder.  So currently we use libcompose in a server side context.  In that context we don't want to allow builds that reference the disk.  So we only allow remote builds if libcompose is ran in a server side context.

vdemeester:
@ibuildthecloud Using it on rancher-compose right ?

I'm wondering something though, should it be there (in `libcompose`) or in `docker/docker` ? I would think in `docker/docker` in a non-internal package (like the api one that should be there at some point or in a `pkg` — but I'm not that sure about `pkg`).

ibuildthecloud:
In docker/docker would be preferred and that is where I copied the list to begin with.  Let me look from where.  I imagine it was private.

vdemeester:
@ibuildthecloud @dnephin I think we should fill an issue for *make this better* and merge this one as it breaks current codes, wdyt ? (the actual list feels weird at it would mean `git@github.com:vdemeester/libcompose` is a valid git+ssh remote but `vincent@myhost:src/docker/libcompose` would not :sweat:  — but that's outside the scope of this PR).

dnephin:
Ya, I'm ok merging this, it shouldn't hurt anything

vdemeester:
Created #95, so LGTM 🐺

PR #74, opened 2015-10-07
Expose Interpolate for external tools

ibuildthecloud:
It is quite useful to parse a text file in the same exact way that
compose does.  One can imagine this being useful for GUI applications or
custom parsers.

### DIFF
--- a/project/interpolation.go
+++ b/project/interpolation.go
@@ -136,7 +136,8 @@ func parseConfig(option, service string, data *interface{}, mapping func(string)
 	return nil
 }
 
-func interpolate(environmentLookup EnvironmentLookup, config *rawServiceMap) error {
+// Interpolate replaces variables in the raw map representation of the project file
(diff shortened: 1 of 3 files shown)


ibuildthecloud:
I started integrating the variable interpolation into some tools and have found that it is much nicer to be able to be able to interpolate the file the same way that libcompose does internally.

vdemeester:
@ibuildthecloud  Seems ok to me :+1: 

One question though : should or shouldn't it be in a separate package (like `libcompose/interpolate` or something like the `pkg/…` package in docker)

ibuildthecloud:
Hmm.  Yes I could move it to libcompose/interpolate.  I like that.

vdemeester:
@ibuildthecloud still working on it or you would like me to carry it ? :wink:

vdemeester:
Carried by @joshwget in #181

PR #76, opened 2015-10-08
Replacing log package with logrus

ibuildthecloud:
The wrong log package was used resulting in odd logging behavior. All
code should use the logrus package (not renamed to log) instead of the
default go log package.


dnephin:
LGTM

PR #77, opened 2015-10-08
Update all target to mimic jenkins

### LINKED ISSUE #75: `make all` doesn't do "all"
I'm not sure how the build system and Jenkins is setup, so I don't fully understand the Makefile setup.  Currently `make all` just does a build and not run tests.  I keep forgetting to run `test` or `validate` before I check in, resulting in failed builds on Jenkins.

Is there a simple target I can run that will run all tests and validation basically simulating what Jenkins would do?

@vdemeester

vdemeester:
Using make.sh without any "bundle" will run the defaults one, just like
on the CI. 🐻

Closes #75

/cc @ibuildthecloud 

🐸

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ DOCKER_RUN_LIBCOMPOSE := docker run --rm -it --privileged $(LIBCOMPOSE_ENVS) $(L
 default: binary
 
 all: build
+	$(DOCKER_RUN_LIBCOMPOSE) ./script/make.sh
 
 binary: build
 	$(DOCKER_RUN_LIBCOMPOSE) ./script/make.sh binary


Makefile
```diff
@@ -18,6 +18,7 @@ DOCKER_RUN_LIBCOMPOSE := docker run --rm -it --privileged $(LIBCOMPOSE_ENVS) $(L
 default: binary
 
 all: build
```
dnephin: Do we still need this build dependency here? I think it'll be run by `make.sh` right ?

vdemeester: @dnephin yes we do, build just build the `build` image (build inception :stuck_out_tongue_closed_eyes:), it doesn't build anything per se (it's `binary` that builds the binary).

*way too many `build` in this sencence*

dnephin: Oh right! ok

ibuildthecloud:
LGTM.  I ran this locally and works as advertised :smile:

PR #81, opened 2015-10-11
Remove docker/docker/{api,graph/tags} dependencies

vdemeester:
Part of #32


ibuildthecloud:
@vdemeester With this PR should Godeps be updated too to trim out the packages we don't need anymore?

vdemeester:
@ibuildthecloud It does, it's just github that doesn't display the `Godep.json` for whatever reason :sweat:.

ibuildthecloud:
@vdemeester (sidenote, godeps is a pain), but if you run `godep restore; rm -rf Godeps; godep save` it should end up deleting a bunch of files.  Right now since the files are in the repo we could be accidentally compiling against them through some indirect reference.

vdemeester:
ahhh right, you meant removing them from the `Godeps/_workspace/src` folder !! Yeah, I'll do it real quick :wink:.

(sidenote, I have the same opinion on `godep`, I'm looking for alternatives :cat:)

vdemeester:
@ibuildthecloud Updated (way more deletion than I though :stuck_out_tongue_closed_eyes:).

dnephin:
I looked around for alternatives to godep (I'm also not a fan). http://jbeckwith.com/2015/05/29/dependency-management-go/ has a long list.  I think the vendor experiment in 1.5 is going to make things a lot better. 

https://github.com/nitrous-io/goop looked good, but hasn't been touched in a year, so probably not a good options

https://github.com/Masterminds/glide looks like really awesome. It even supports using forks of dependencies, and removes the need to check in all your dependencies!

ibuildthecloud:
LGTM

ibuildthecloud:
Something based on 1.5 vendoring would be good, but I don't have opinions on which system though.  But that does raise the point that we should probably bump libcompose to go 1.5.

vdemeester:
@ibuildthecloud true, I'll make a PR for it :wink:.

@ibuildthecloud @dnephin on alternative, to make it simple, I do think we should use something based on 1.5 vendoring as this could/should be the way to go in the golang community (and thus would be straightforward for any dev' to *get in*).

dnephin:
Cool, so https://github.com/Masterminds/glide is based on the 1.5 vendoring. I noticed that the docs for the 1.5 vendoring say that they're only designed to solve the import problem. It's still up to community tooling to support the fetching of dependencies, etc.

So `glide` is one version of that tooling that seems to have a decent following and active development already.

dnephin:
LGTM

PR #83, opened 2015-10-13
Bump go version to 1.5

vdemeester:
To discuss :cat:.

- Bump to Go 1.5.
- Use `GO15VENDOREXPERIMENT` (and `glide` but could be anything else.. still experimenting with something else though).

The idea behind it is to not even need `glide` for building the project (if sources are vendored). `glide` would only be needed for managing dependencies (updating them, adding/removing them, …).

🐸


vdemeester:
Hum, there is a weird test failure.. :cry:.
I also wonder if I could do the move to `glide` here or in a separate PR (already done some work locally in another branch)

ibuildthecloud:
Let me check out glide, I'm not familiar with that tool.  In other projects we have found that 1.5 vendoring to be pretty ruff around the edges.  For example `go test ./...` runs all vendored tests.

vdemeester:
@ibuildthecloud yeah, I'm looking at a different way to handle that, more like on `docker/docker` (to remove vendored test & co).

dnephin:
>  For example go test ./... runs all vendored tests.

https://github.com/Masterminds/glide#glide-novendor-aliased-to-nv addresses that problem

I've started to use glide for a personal project. So far I like the design decisions they've made.  There are some rough edges (since it's fairly new), but it sounds like they have a bunch of WIP that will smooth out those problems.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # This file describes the standard way to build libcompose, using docker
-FROM golang:1.4.2-cross
+FROM golang:1.5
```
imikushin: Why not use 1.5.1?

imikushin:
:+1: for go-1.5+ and vendoring.

Dockerfile
vdemeester: Hum… good point :stuck_out_tongue_closed_eyes:

imikushin:
Looking forward to have it merged!

vdemeester:
I kinda like `glide` but would really like to have some "clean the mess" script as we do in `docker/docker` to remove tests, `Godeps/vendor` folder & co from dependencies we have — so that the checkout is lightweight.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # This file describes the standard way to build libcompose, using docker
-FROM golang:1.4.2-cross
+FROM golang:1.5.1
```
dnephin: I think `golang:1.5` was actually the better way to go. If you look at https://hub.docker.com/_/golang/ they're both the same image right now. But if 1.5.2 gets released, we'd be auto-upgraded to use it.

Unless we expect a minor release to break something, it's probably easier to go with the more general tag.

This is minor though, I'm ok either way.

dnephin:
+1 for removing `vendor/` from git.  Managing dependencies in git is the worst part of golang.  Could you do that with this change instead of re-adding them all under a different name?

Dockerfile
vdemeester: I wondered a bit about that, but the more explicit we are, the more *sure* we are that we build the same thing ; using `1.5` only would mean some could build it using 1.5.1 and the other 1.5.2, etc.. but as you said, I don't think we expect it to break with minor release so I'm fine with both :stuck_out_tongue:.

vdemeester:
I'm 50/50 about the vendoring but.. meh.. I'll try without it to :)

imikushin:
@vdemeester would be great to add `/vendor` to .gitignore then

vdemeester:
True, I'll try to update the PR this week (rebasing, squashing and small updates :wink:)

imikushin:
I'd really like godep to `go away`

vdemeester:
Rebased, squashed and updated.

- Removed Godeps folder altogether, this means no more vendoring.
- Setup a `glide.yaml` from the old `Godeps.json` (to keep the same version of deps).
- Updated the `Dockerfile` to run a `glide up` to get dependencies.
- Fix an integration that fails :sweat:

dnephin:
Removing the deps looks good, just a question about the test change

imikushin:
@vdemeester This file is in the vendor dir after the checkout: `./vendor/github.com/docker/docker/autogen/dockerversion/dockerversion.go`

Looks like it's been forgotten here alone :smiley_cat:

vdemeester:
@imikushin this is related to that : https://github.com/docker/docker/issues/11699 ; It's gonna be fixed at some point https://github.com/docker/docker/pull/17431 (it's a matter of days). It has not been forgotten and left there for now — although we could find some way to generate it like in `docker/docker` for now or with `go generate` for the time being.

@ibuildthecloud @dnephin What do you think about that ? (using `go generate` would felt more `go`-friendly)

imikushin:
Would it be a good idea to add `vendor` to .dockerignore as well? 
Doesn't `glide up` have to download the sources regardless of if they are present already in `./vendor`?

imikushin:
@vdemeester Nevermind my previous comment: vendor shouldn't be in .dockerignore (yet) because it has that file.

vdemeester:
@imikushin exactly :stuck_out_tongue_closed_eyes:

imikushin:
I wonder where it caches the docker packages in the container...

Because there's "that file" I can't build or test locally: `glide up` warns about it:
```
[WARN] github.com/docker/docker appears to be a vendored package. Unable to update. Consider the '--update-vendored' flag.
```

Then `go test github.com/docker/libcompose/docker` can't find docker packages:
```
go test ./docker
docker/client.go:10:2: cannot find package "github.com/docker/docker/cliconfig" in any of:
        /Users/ivan/src/docker/libcompose/src/github.com/docker/libcompose/vendor/github.com/docker/docker/cliconfig (vendor tree)
        /usr/local/Cellar/go/1.5.1/libexec/src/github.com/docker/docker/cliconfig (from $GOROOT)
        /Users/ivan/src/docker/libcompose/src/github.com/docker/docker/cliconfig (from $GOPATH)
docker/client.go:11:2: cannot find package "github.com/docker/docker/opts" in any of:
        /Users/ivan/src/docker/libcompose/src/github.com/docker/libcompose/vendor/github.com/docker/docker/opts (vendor tree)
        /usr/local/Cellar/go/1.5.1/libexec/src/github.com/docker/docker/opts (from $GOROOT)
        /Users/ivan/src/docker/libcompose/src/github.com/docker/docker/opts (from $GOPATH)
docker/builder.go:13:2: cannot find package "github.com/docker/docker/pkg/archive" in any of:
        /Users/ivan/src/docker/libcompose/src/github.com/docker/libcompose/vendor/github.com/docker/docker/pkg/archive (vendor tree)
        /usr/local/Cellar/go/1.5.1/libexec/src/github.com/docker/docker/pkg/archive (from $GOROOT)
        /Users/ivan/src/docker/libcompose/src/github.com/docker/docker/pkg/archive (from $GOPATH)
docker/builder.go:14:2: cannot find package "github.com/docker/docker/pkg/fileutils" in any of:
        /Users/ivan/src/docker/libcompose/src/github.com/docker/libcompose/vendor/github.com/docker/docker/pkg/fileutils (vendor tree)
        /usr/local/Cellar/go/1.5.1/libexec/src/github.com/docker/docker/pkg/fileutils (from $GOROOT)
        /Users/ivan/src/docker/libcompose/src/github.com/docker/docker/pkg/fileutils (from $GOPATH)
docker/client.go:12:2: cannot find package "github.com/docker/docker/pkg/homedir" in any of:
        /Users/ivan/src/docker/libcompose/src/github.com/docker/libcompose/vendor/github.com/docker/docker/pkg/homedir (vendor tree)
        /usr/local/Cellar/go/1.5.1/libexec/src/github.com/docker/docker/pkg/homedir (from $GOROOT)
        /Users/ivan/src/docker/libcompose/src/github.com/docker/docker/pkg/homedir (from $GOPATH)
docker/convert.go:6:2: cannot find package "github.com/docker/docker/pkg/nat" in any of:
        /Users/ivan/src/docker/libcompose/src/github.com/docker/libcompose/vendor/github.com/docker/docker/pkg/nat (vendor tree)
        /usr/local/Cellar/go/1.5.1/libexec/src/github.com/docker/docker/pkg/nat (from $GOROOT)
        /Users/ivan/src/docker/libcompose/src/github.com/docker/docker/pkg/nat (from $GOPATH)
docker/container.go:10:2: cannot find package "github.com/docker/docker/pkg/parsers" in any of:
        /Users/ivan/src/docker/libcompose/src/github.com/docker/libcompose/vendor/github.com/docker/docker/pkg/parsers (vendor tree)
        /usr/local/Cellar/go/1.5.1/libexec/src/github.com/docker/docker/pkg/parsers (from $GOROOT)
        /Users/ivan/src/docker/libcompose/src/github.com/docker/docker/pkg/parsers (from $GOPATH)
docker/client.go:13:2: cannot find package "github.com/docker/docker/pkg/tlsconfig" in any of:
        /Users/ivan/src/docker/libcompose/src/github.com/docker/libcompose/vendor/github.com/docker/docker/pkg/tlsconfig (vendor tree)
        /usr/local/Cellar/go/1.5.1/libexec/src/github.com/docker/docker/pkg/tlsconfig (from $GOROOT)
        /Users/ivan/src/docker/libcompose/src/github.com/docker/docker/pkg/tlsconfig (from $GOPATH)
docker/container.go:11:2: cannot find package "github.com/docker/docker/registry" in any of:
        /Users/ivan/src/docker/libcompose/src/github.com/docker/libcompose/vendor/github.com/docker/docker/registry (vendor tree)
        /usr/local/Cellar/go/1.5.1/libexec/src/github.com/docker/docker/registry (from $GOROOT)
        /Users/ivan/src/docker/libcompose/src/github.com/docker/docker/registry (from $GOPATH)
docker/convert.go:7:2: cannot find package "github.com/docker/docker/runconfig" in any of:
        /Users/ivan/src/docker/libcompose/src/github.com/docker/libcompose/vendor/github.com/docker/docker/runconfig (vendor tree)
        /usr/local/Cellar/go/1.5.1/libexec/src/github.com/docker/docker/runconfig (from $GOROOT)
        /Users/ivan/src/docker/libcompose/src/github.com/docker/docker/runconfig (from $GOPATH)
docker/builder.go:15:2: cannot find package "github.com/docker/docker/utils" in any of:
        /Users/ivan/src/docker/libcompose/src/github.com/docker/libcompose/vendor/github.com/docker/docker/utils (vendor tree)
        /usr/local/Cellar/go/1.5.1/libexec/src/github.com/docker/docker/utils (from $GOROOT)
        /Users/ivan/src/docker/libcompose/src/github.com/docker/docker/utils (from $GOPATH)
```

vdemeester:
@imikushin yeah that's the problem of the file :sob:. It works on the `make` target because we do a `glide up` without it and then we copy this file inside.

So locally it makes things way harder to manage. I'm gonna update this PR to use `go generate` then. The workflow to build it locally without docker will be :

```bash
$ glide up
$ go generate
$ go build
```

Or we wait for #17431 to be merged (re-merged :sweat_smile:) and bump the docker version.

imikushin:
Shoot! I'm an idiot. Of course, `COPY .` is the last operation in Dockerfile.

I'm fine with the `go generate` workflow both locally and in the make targets.

vdemeester:
@imikushin Should be ok now.. I think my *generate* is a little bit harsh (might not work on Windows at first :stuck_out_tongue_closed_eyes:) but *meh* for now :sweat_smile:

imikushin:
Perfect! Well, almost:  now adding `vendor` to .dockerignore makes total sense, huh? :wink:

vdemeester:
And it's done `\o/`

imikushin:
LGTM :fireworks:

dnephin:
LGTM, happy to see the deps removed from git!

vdemeester:
Hum, I think I should update the `README` too (about the `go generate`)

imikushin:
Nice! 
@ibuildthecloud merge? :wink:

ibuildthecloud:
(A very strong) -1 on removing vendor/.  The git world is not well managed in terms of publishing code.  In other ecosystems like Python and Java there are reliable repositories for finding software. Go you just grab the source from some random git URL.  That git could go away, change, be deleted, rebased, etc.  It is far better to copy the source into your project.  Then you know you can always reproduce the build.

ibuildthecloud:
Not LGTM

vdemeester:
@ibuildthecloud as said, I'm 50/50 on the vendoring stuff, so it feels fine to me to vendor them too :wink:. If we keep sources in the repo though, I'm thinking of "ditching" `glide` and use a script like on `docker/docker` (that removes some things like tests files to keep the repos a little cleaner). WDYT ?

dnephin:
I think if we're using a dependency that is just going to disappear, we should not be using that library at all.  

Looking at our dependencies right now, most urls are github, `golang.org`, or `gopkg.in`.  `gopkg.in` actually seems like it acts as that service for go, since it would allow you to move the underlying repo to a new url (I think).

Do you think there is actually a reasonable chance that one of our dependencies gets removed?

ibuildthecloud:
@dnephin There's no reason to go against the accepted approach in go.  This is why vendor exists to begin with.  For example, one major reason of `vendor/` is to allow your code to be `go get`-table but then still maintain your specific library dependencies.  Without vendoring the code we can't support that because you then have to assume your code always works with everybody's master.

It is very well accepted and discussed in the go community that vendoring is a best practice.  I don't think we need to deviate.

ibuildthecloud:
@dnephin Also realize that gopkg.in is really just some dude at the end of the day.

ibuildthecloud:
@vdemeester I'm fine with ditching glide too.  If `docker/docker` has a good script we can use I'm all for it.  Basically all our scripts have been ~~ripped off~~ borrowed from docker/docker :smile:

vdemeester:
@ibuildthecloud well yeah :sweat_smile:.

dnephin:
>  Without vendoring the code we can't support that because you then have to assume your code always works with everybody's master.

That's not true at all. You still pin a specific git sha.  The only difference is in how dependencies are included. Either you keep a copy locally, or you run `glide up` to fetch them.  You end up with the same source files.

imikushin:
I'm with @dnephin on this one, but whatever... Basically just want to use go1.5 built-in vendored package resolution.

vdemeester:
And it's up-to-date. [vendor.sh](https://github.com/vdemeester/libcompose/blob/bump-go-1.5/script/vendor.sh) and [.vendor-helpers.sh](https://github.com/vdemeester/libcompose/blob/bump-go-1.5/script/.vendor-helpers.sh) do quite a good job to clean un-necessary files.

imikushin:
Let me take it for a spin real quick.

imikushin:
README.md:
`go build -o libcompose cli/main` doesn't work 
("cli/main" not found neither in $GOPATH, nor in $GOROOT)

`go build -o libcompose ./cli/main` does

vdemeester:
Ah you're right :sweat_smile:, will update :wink:

imikushin:
Looks good. 
Ugly, but as good as vendoring can look :smiling_imp:

ibuildthecloud:
That's actually a pretty clean vendor directory.  I'm impressed with that script, props to @tianon.  Even maintains the LICENSE and COPYING files.

But I'm just going to pretend the `go generate` doesn't exist :)

LGTM

@dnephin I'd like to wait for a LGTM from you in case you have further comments before we merge.

vdemeester:
> But I'm just going to pretend the `go generate` doesn't exist :)

It should exists for a short period of time :wink:.

dnephin:
How do you review these PRs when github doesn't show you the full change because there are too many files being changed?

I guess I have to get a checkout of this branch and craft the right git command to show me the diff without the Godeps and vendor directories?

dnephin:
I guess it's better than Godeps... so let's do it

PR #86, opened 2015-10-16
Strings are valid for numeric fields

### LINKED ISSUE #85: Numeric attributes should also be accepted as strings
Docker Compose accepts strings for all attributes that accept numbers. For example, these both should be valid:

    mem_limit: 40000000
    mem_limit: "40000000"

Currently, only the former is accepted by libcompose. This also breaks interpolation since variables are always inserted as strings.

One possible solution is to change the types from int64 to the following:

    type Stringorint64 struct {
        value int64
    }

    func (s Stringorint64) MarshalYAML() (interface{}, error) {
        return s.value, nil
    }

    func (s *Stringorint64) UnmarshalYAML(unmarshal func(interface{}) error) error {
        var str string
        err := unmarshal(&str)

        if err != nil {
            return err
        }

        i, err := strconv.ParseInt(str, 10, 64)

        if err != nil {
            return err
        }

        s.value = i

        return nil
    }

    func (s *Stringorint64) Value() int64 {
        return s.value
    }

This works, but it's a little cumbersome to have to use `Value()` to access those attributes. Any better ideas to go about this?

joshwget:
Fixes #85. This works by attempting to convert all fields to integers. If it's successful for a given field, it sets the type as an integer, and if not it sets it as a string.

Then, if the type of that field should be an integer in `ServiceConfig`, all is well. If not, then go-yaml will convert that field back into a string when it unmarshals to a  `ServiceConfig`.

Thanks to @ibuildthecloud for thinking of this method.

### DIFF
--- a/integration/create_test.go
+++ b/integration/create_test.go
@@ -2,8 +2,9 @@ package integration
 
 import (
 	"fmt"
-	. "gopkg.in/check.v1"
 	"os"
+
+	. "gopkg.in/check.v1"
 )
 
 func (s *RunSuite) TestFields(c *C) {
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
LGTM

vdemeester:
LGTM 🐼

PR #87, opened 2015-10-17
Environment variables can contain equal signs

joshwget:
Environment variables containing an equals sign weren't being parsed correctly. This is a fix for that issue, as well as an additional unit test to cover that case.

### DIFF
--- a/project/interpolation.go
+++ b/project/interpolation.go
@@ -147,7 +147,12 @@ func interpolate(environmentLookup EnvironmentLookup, config *rawServiceMap) err
 					return ""
 				}
 
-				return strings.Split(values[0], "=")[1]
+				// Use first result if many are given
+				value := values[0]
+
+				// Environment variables come in key=value format
+				// Return everything past first '='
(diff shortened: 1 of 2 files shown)


project/interpolation.go
```diff
@@ -147,7 +147,12 @@ func interpolate(environmentLookup EnvironmentLookup, config *rawServiceMap) err
 					return ""
 				}
 
-				return strings.Split(values[0], "=")[1]
+				// Use first result if many are given
+				value := values[0]
+
+				// Environment variables come in key=value format
+				// Return everything past first '='
+				return strings.Join(strings.Split(value, "=")[1:], "=")
```
ibuildthecloud: I think what you want here is SplitN(value, "=", 2)[1]

joshwget: Yeah that's definitely better here

ibuildthecloud:
LGTM

vdemeester:
LGTM 🐰

PR #89, opened 2015-10-19
Support recreating containers

imikushin:
In lieu of #16

### DIFF
--- a/cli/command/command.go
+++ b/cli/command/command.go
@@ -72,8 +72,12 @@ func UpCommand(factory app.ProjectFactory) cli.Command {
 				Usage: "Do not block and log",
 			},
 			cli.BoolFlag{
-				Name:  "rebuild",
-				Usage: "Rebuild containers",
+				Name:  "no-recreate",
+				Usage: "If containers already exist, don't recreate them. Incompatible with --force-recreate.",
+			},
+			cli.BoolFlag{
+				Name:  "force-recreate",
+				Usage: "Recreate containers even if their configuration and image haven't changed. Incompatible with --no-recreate.",
 			},
 		},
 	}
@@ -231,7 +235,8 @@ func Populate(context *project.Context, c *cli.Context) {
 		context.Log = true
 	} else if c.Command.Name == "up" {
 		context.Log = !c.Bool("d")
-		context.Rebuild = c.Bool("rebuild")
+		context.NoRecreate = c.Bool("no-recreate")
+		context.ForceRecreate = c.Bool("force-recreate")
 	} else if c.Command.Name == "stop" || c.Command.Name == "restart" || c.Command.Name == "scale" {
 		context.Timeout = uint(c.Int("timeout"))
 	} else if c.Command.Name == "kill" {
--- a/cli/command/command.go
+++ b/cli/command/command.go
@@ -71,6 +71,10 @@ func UpCommand(factory app.ProjectFactory) cli.Command {
 				Name:  "d",
 				Usage: "Do not block and log",
 			},
+			cli.BoolFlag{
+				Name:  "rebuild",
+				Usage: "Rebuild containers",
+			},
 		},
 	}
 }
@@ -227,6 +231,7 @@ func Populate(context *project.Context, c *cli.Context) {
 		context.Log = true
 	} else if c.Command.Name == "up" {
 		context.Log = !c.Bool("d")
+		context.Rebuild = c.Bool("rebuild")
 	} else if c.Command.Name == "stop" || c.Command.Name == "restart" || c.Command.Name == "scale" {
 		context.Timeout = uint(c.Int("timeout"))
 	} else if c.Command.Name == "kill" {
--- a/docker/container.go
+++ b/docker/container.go
@@ -13,6 +13,7 @@ import (
 	"github.com/docker/docker/utils"
 	"github.com/docker/libcompose/logger"
 	"github.com/docker/libcompose/project"
+	util "github.com/docker/libcompose/utils"
 	dockerclient "github.com/fsouza/go-dockerclient"
 )
 
@@ -100,7 +101,8 @@ func name(names []string) string {
 	return current[1:]
 }
 
-func (c *Container) Rebuild(imageName string) (*dockerclient.Container, error) {
+// Rebuild will delete the container and create a new one with the current configuration
+func (c *Container) Rebuild(imageName string) (*dockerclient.APIContainers, error) {
 	info, err := c.findInfo()
 	if err != nil {
 		return nil, err
@@ -114,15 +116,11 @@ func (c *Container) Rebuild(imageName string) (*dockerclient.Container, error) {
 	}
 
 	name := info.Name[1:]
-	new_name := fmt.Sprintf("%s_%s", name, info.ID[:12])
-	deleted := false
-	logrus.Debugf("Renaming %s => %s", name, new_name)
-	if err := c.client.RenameContainer(name, new_name); err != nil {
-		logrus.Errorf("Rename failed, deleting %s", name)
-		if err := c.client.RemoveContainer(info.ID, true, false); err != nil {
-			return nil, err
-		}
-		deleted = true
+	newName := fmt.Sprintf("%s_%s", name, info.ID[:12])
+	logrus.Debugf("Renaming %s => %s", name, newName)
+	if err := c.client.RenameContainer(dockerclient.RenameContainerOptions{ID: info.ID, Name: newName}); err != nil {
+		logrus.Errorf("Failed to rename old container %s", c.name)
+		return nil, err
 	}
 
 	newContainer, err := c.createContainer(imageName, info.ID)
@@ -131,13 +129,13 @@ func (c *Container) Rebuild(imageName string) (*dockerclient.Container, error) {
 	}
 	logrus.Debugf("Created replacement container %s", newContainer.ID)
 
-	if !deleted {
-		if err := c.client.RemoveContainer(info.ID, true, false); err != nil {
-			logrus.Errorf("Failed to remove old container %s", c.name)
-			return nil, err
-		}
-		logrus.Debugf("Removed old container %s %s", c.name, info.ID)
+	if err := c.client.RemoveContainer(
+		dockerclient.RemoveContainerOptions{ID: info.ID, Force: true, RemoveVolumes: false}); err != nil {
+
+		logrus.Errorf("Failed to remove old container %s", c.name)
+		return nil, err
 	}
+	logrus.Debugf("Removed old container %s %s", c.name, info.ID)
 
 	return newContainer, nil
 }
@@ -224,13 +222,9 @@ func (c *Container) Up(imageName string) error {
 	}
 
 	if !info.State.Running {
-		logrus.Debugf("Starting container: %s: %#v", container.ID, info.HostConfig)
-		err = c.populateAdditionalHostConfig(info.HostConfig)
-		if err != nil {
-			return err
-		}
-
-		if err := c.client.StartContainer(container.ID, info.HostConfig); err != nil {
+		logrus.WithFields(logrus.Fields{"container.ID": container.ID, "c.name": c.name}).Debug("Starting container")
+		if err = c.client.StartContainer(container.ID, nil); err != nil {
+			logrus.WithFields(logrus.Fields{"container.ID": container.ID, "c.name": c.name}).Debug("Failed to start container")
 			return err
 		}
 
@@ -273,7 +267,17 @@ func (c *Container) OutOfSync(imageName string) (bool, error) {
 }
 
 func (c *Container) getHash() string {
-	return project.GetServiceHash(c.service.Name(), *c.service.Config())
+	return project.GetServiceHash(c.service.Name(), c.service.Config())
+}
+
+func volumeBinds(volumes map[string]struct{}, container *dockerclient.Container) []string {
+	result := make([]string, 0, len(container.Mounts))
+	for _, mount := range container.Mounts {
+		if _, ok := volumes[mount.Destination]; ok {
+			result = append(result, fmt.Sprint(mount.Source, ":", mount.Destination))
+		}
+	}
+	return result
 }
 
 func (c *Container) createContainer(imageName, oldContainer string) (*dockerclient.APIContainers, error) {
@@ -299,18 +303,22 @@ func (c *Container) createContainer(imageName, oldContainer string) (*dockerclie
 	}
 
 	if oldContainer != "" {
-		createOpts.HostConfig.VolumesFrom = append(createOpts.HostConfig.VolumesFrom, oldContainer)
+		info, err := c.client.InspectContainer(oldContainer)
+		if err != nil {
+			return nil, err
+		}
+		createOpts.HostConfig.Binds = util.Merge(createOpts.HostConfig.Binds, volumeBinds(createOpts.Config.Volumes, info))
 	}
 
 	logrus.Debugf("Creating container %s %#v", c.name, createOpts)
 
-	id, err := c.client.CreateContainer(*createOpts)
+	container, err := c.client.CreateContainer(*createOpts)
 	if err != nil && err == dockerclient.ErrNoSuchImage {
 		logrus.Debugf("Not Found, pulling image %s", createOpts.Config.Image)
 		if err = c.pull(createOpts.Config.Image); err != nil {
 			return nil, err
 		}
-		if id, err = c.client.CreateContainer(*createOpts); err != nil {
+		if container, err = c.client.CreateContainer(*createOpts); err != nil {
 			return nil, err
 		}
 	}
@@ -320,7 +328,7 @@ func (c *Container) createContainer(imageName, oldContainer string) (*dockerclie
 		return nil, err
 	}
 
-	return GetContainerById(c.client, id)
+	return GetContainerByID(c.client, container.ID)
 }
 
 func (c *Container) populateAdditionalHostConfig(hostConfig *dockerclient.HostConfig) error {
@@ -460,10 +468,10 @@ func (c *Container) Log() error {
 }
 
 func (c *Container) pull(image string) error {
-	return PullImage(c.client, c.service, image)
+	return pullImage(c.client, c.service, image)
 }
 
-func PullImage(client dockerclient.Client, service *Service, image string) error {
+func pullImage(client *dockerclient.Client, service *Service, image string) error {
 	taglessRemote, tag := parsers.ParseRepositoryTag(image)
 	if tag == "" {
 		image = utils.ImageReference(taglessRemote, DefaultTag)
--- a/docker/container.go
+++ b/docker/container.go
@@ -101,13 +101,14 @@ func name(names []string) string {
 	return current[1:]
 }
 
-// Rebuild will delete the container and create a new one with the current configuration
-func (c *Container) Rebuild(imageName string) (*dockerclient.APIContainers, error) {
+// Recreate will not refresh the container by means of relaxation and enjoyment,
+// just delete it and create a new one with the current configuration
+func (c *Container) Recreate(imageName string) (*dockerclient.APIContainers, error) {
 	info, err := c.findInfo()
 	if err != nil {
 		return nil, err
 	} else if info == nil {
-		return nil, fmt.Errorf("Can not find container to rebuild for service: %s", c.service.Name())
+		return nil, fmt.Errorf("Can not find container to recreate for service: %s", c.service.Name())
 	}
 
 	hash := info.Config.Labels[HASH.Str()]
--- a/docker/container.go
+++ b/docker/container.go
@@ -3,6 +3,7 @@ package docker
 import (
 	"fmt"
 	"math"
+	"os"
 	"strings"
 
 	"github.com/Sirupsen/logrus"
@@ -13,7 +14,6 @@ import (
 	"github.com/docker/libcompose/logger"
 	"github.com/docker/libcompose/project"
 	dockerclient "github.com/fsouza/go-dockerclient"
-	"os"
 )
 
 // DefaultTag is the name of the default tag of an image.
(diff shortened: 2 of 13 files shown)


imikushin:
I've addressed the review comments, and there's one "FIXME" left to discuss. When that's addressed, this PR will be good to merge.

ibuildthecloud:
@imikushin We need to do volumes from the old container so that it copies all the same mounts as the older container.  Then we delete the container later.  I believe the behavior of volumes from is that is copies the info from the old container and then you can delete the old container.

vdemeester:
This is how compose works I think. /cc @dnephin

dnephin:
We don't use `volume_from` any more because it ends up copying over some volumes we don't want. The new logic is:

* rename old container to a temporary name
* create a new container with the old volumes (https://github.com/docker/compose/blob/5ed9f9/compose/service.py#L897-L910)
* remove old container with temporary name

So we pick out only the data volumes that are still in the new configuration, and copy those over.

ibuildthecloud:
@dnephin This is close to what we do now in this PR, but we are still using `volumes-from`.  @imikushin Can you do the same as docker-compose?  https://github.com/docker/compose/blob/5ed9f9/compose/service.py#L913-L938  That should be a simple change but we I believe we already to the renaming logic.

imikushin:
I've implemented mounts migration.

imikushin:
Updated volume migration logic and added a test. @ibuildthecloud @dnephin @vdemeester Please take a look.

docker/container.go
```diff
@@ -198,7 +240,7 @@ func (c *Container) Up(imageName string) error {
 
 // OutOfSync checks if the container is out of sync with the service definition.
 // It looks if the the service hash container label is the same as the computed one.
-func (c *Container) OutOfSync() (bool, error) {
+func (c *Container) OutOfSync(imageName string) (bool, error) {
```
ibuildthecloud: Is imageName used?

docker/container.go
```diff
@@ -209,10 +251,28 @@ func (c *Container) OutOfSync() (bool, error) {
...
+	}).Debug("Comparing old and must-be hash")
+	return info.Config.Labels[HASH.Str()] != c.getHash(), nil
+}
+
+func (c *Container) getHash() string {
+	return project.GetServiceHash(c.service.Name(), c.service.Config())
 }
 
-func (c *Container) createContainer(imageName string) (*dockerclient.APIContainers, error) {
+func volumeBinds(volumes map[string]struct{}, container *dockerclient.Container) []string {
```
ibuildthecloud: The python code looks at external: https://github.com/docker/compose/blob/5ed9f9/compose/service.py#L926  Is that covered here?

docker/container.go
imikushin: oops! not anymore

docker/container.go
imikushin: Yeah, the `volumes` arg here doesn't have external volumes, so no need to check.

ibuildthecloud:
LGTM

ibuildthecloud:
@vdemeester I'm particularly interested in getting this PR merged soon.  We really need this functionality so currently we have a rancher specific fork of libcompose that we ship in Rancher.  We really want to move to upstream but really haven't been able to because of this PR.

There's probably bugs in this or something isn't 100% correct, but I'd rather merge the PR and then iterate on bugs then wait for something that is perfect.

docker/container.go
```diff
@@ -96,6 +97,51 @@ func name(names []string) string {
 	return current[1:]
 }
 
+// Rebuild will delete the container and create a new one with the current configuration
+func (c *Container) Rebuild(imageName string) (*dockerclient.APIContainers, error) {
```
dnephin: I think "build" is the wrong terminology here. We build images, but we create containers. I think it should be `Recreate()`

dnephin: same thing with the error messages below

vdemeester:
I'm ok with, reviewing tomorrow and we'll get this merged soon ;-)

docker/container.go
```diff
@@ -96,6 +97,51 @@ func name(names []string) string {
...
+
+	name := info.Name[1:]
+	newName := fmt.Sprintf("%s_%s", name, info.ID[:12])
+	deleted := false
+	logrus.Debugf("Renaming %s => %s", name, newName)
+	if err := c.client.RenameContainer(dockerclient.RenameContainerOptions{ID: info.ID, Name: newName}); err != nil {
+		logrus.Errorf("Rename failed, deleting %s", name)
+		if err := c.client.RemoveContainer(
+			dockerclient.RemoveContainerOptions{ID: info.ID, Force: true, RemoveVolumes: false}); err != nil {
+			return nil, err
```
dnephin: This is wrong. If rename fails you can't remove the old container. That would cause you to lose your volumes (unless you go in and manually copy them over somewhere, which wasn't possible before 1.9).

docker/service.go
```diff
@@ -186,15 +200,59 @@ func (s *Service) up(imageName string, create bool) error {
...
+		"newRebuildLabel":     newRebuildLabel,
+		"outOfSync":           outOfSync,
+		"ForceRecreate":       s.context.ForceRecreate,
+		"rebuildLabelChanged": rebuildLabelChanged}).Debug("Rebuild values")
+
+	if s.context.ForceRecreate || origRebuildLabel == "always" || rebuildLabelChanged || origRebuildLabel != "false" && outOfSync {
+		logrus.Infof("Rebuilding %s", name)
+		if _, err := c.Rebuild(imageName); err != nil {
+			return err
+		}
```
dnephin: This doesn't look like any logic we have in docker-compose, what's this about?

dnephin: I don't see where this `REBUILD` label is being set on the container.

Why does it matter what the flag was when we originally created the container? I think the only thing that matters is the current flags to `up`.

dnephin:
@ibuildthecloud  I don't think this is ready to be merged. I've left some initial comments

docker/container.go
vdemeester: Completely agree :+1:

docker/service.go
ibuildthecloud: @dnephin I guess we need to find a different approach then.  So this code was originally taken from RancherOS.  In RancherOS we have containers that must always be recreated on "up".  This is what this logic is about.  By adding a label equal to "always" it basically does force recreate on every up for certain containers.

ibuildthecloud: @dnephin Actually let me explain this logic a bit more.  There are two values you can put for the label REBUILD=always or REBUILD=false.  The context here is specially the "console" container in RancherOS.  The way RancherOS works is that the when the OS boots we essentially do "up".  We have three modes for the console: ephemeral, persistant, "default".

1. ephemeral: An every up we always recreate, the is the equivalent of `--force-recreate`.
1. persistant: Never recreate the container regardless if the compose file changes.  Basically the equivalent of `--no-recreate`
1. default: just normal logic

The logic gets confusing when you recreate container with different values of this REBUILD tag.  For example if REBUILD=false and then you edit the compose file to REBUILD=always or vice versa.  I'll tell you we tested this extensively.

We really need this logic, so if you don't want it in libcompose we will just have to find away of pulling it out.  The key thing here is that we need to essentially set `--force-recreate/--no-recreate` on a specific service in the project and ignore the which mode is set for the current up.

Your call.  I see two paths forward.  We can keep the feature but change the label and terminology to fall more inline with docker-compose, so RECREATE=no or RECREATE=force.  Or we yank it out and just do it in RancherOS.  I think it's a useful feature.

docker/container.go
ibuildthecloud: Makes sense.  @imikushin Can you rename everything to be recreate.  I just messed that one up.

docker/container.go
ibuildthecloud: Okay, originally the rename was done as "best effort" but I agree that you could loose data.  @imikushin Can you just remove this?  It actually makes the code much cleaner.  Just fail if we can't rename.

ibuildthecloud:
@dnephin Thanks for reviewing.  Just to clarify I don't want to rush this in especially if its wrong.  I just was just saying if maybe a corner case or something wasn't covered we should consider merging.  But you right that there as definitely some wrong things here.

imikushin:
Hopefully, now it's less wrong :)

docker/service.go
dnephin: Interesting idea. I agree it does get pretty confusing when you have different containers in a stack that have different recreate policies. I can see a scenario that is just impossible to resolve (at least with links or volumes) where a service is set to RECREATE=no and it has a dependency that is set to RECREATE=always. 

The links will be broken every time, so the RECREATE=no is basically invalid, but a lot of work to detect.

I'd feel a lot better leaving this out of libcompose (at least for now) because it can lead to these kind of impossible scenarios.  With the new parameters I think it should be possible to instrument this from outside of libcompose by setting the label in the config, and inspecting the containers before calling recreate.

project/hash.go
```diff
@@ -11,13 +11,11 @@ import (
...
-// have changed and needs to be rebuild.
-func GetServiceHash(service Service) string {
+// have changed and needs to be recreate.
+func GetServiceHash(name string, config *ServiceConfig) string {
 	hash := sha1.New()
 
-	io.WriteString(hash, fmt.Sprintln(service.Name()))
-
-	config := service.Config()
+	io.WriteString(hash, name)
```
dnephin: I think the config has is missing `ImageId` (it needs to resolve the current id from the name), and we don't track `name` in the hash (any change in name will already be reflected in the labels).

cli/command/command.go
```diff
@@ -221,6 +229,8 @@ func Populate(context *project.Context, c *cli.Context) {
 		context.Log = true
 	} else if c.Command.Name == "up" {
 		context.Log = !c.Bool("d")
+		context.NoRecreate = c.Bool("no-recreate")
+		context.ForceRecreate = c.Bool("force-recreate")
```
dnephin: I think these should be [converted to an enum](https://github.com/docker/compose/blob/f3d49e38c0e488194cd507b092111e0bce21e779/compose/cli/main.py#L622) and passed as a parameter to `project.Up()`. They aren't really part of the "project context", they're a parameter for a single operation.

ibuildthecloud: While your comment is most likely valid, I think it's out of the scope of this PR.  Basically all options from the CLI are currently set in the Context and not passed as options.  We can make that change, but I don't think it's needed for this PR.

project/hash.go
ibuildthecloud: @dnephin Your right, @imikushin something happened here.  In my original PR getHash() took an image name.  You seem to have dropped that, so there is not more logic that checks the imageId has changed.  https://github.com/docker/libcompose/pull/16/files#diff-caac1491e4add415147af39b799d3bdaR258

ibuildthecloud: @imikushin Okay, I figure out what's going on there.  I screwed up.  My original PR had the wrong commit, it wasn't the same logic we are using in RancherOS.  I've cherry picked the right commit and updated #16.  Sorry about that, can you rebase with #16?

imikushin:
Re-rebased @ibuildthecloud 's PR #16 latest incarnation, renamed according to the canon. 
OMG, tests just passed! Please review. 
/cc @vdemeester @dnephin

docker/labels.go
```diff
@@ -11,10 +11,11 @@ type Label string
...
 const (
-	NAME    = Label("io.docker.compose.name")
-	PROJECT = Label("io.docker.compose.project")
-	SERVICE = Label("io.docker.compose.service")
-	HASH    = Label("io.docker.compose.config-hash")
+	NAME     = Label("io.docker.compose.name")
+	PROJECT  = Label("io.docker.compose.project")
+	SERVICE  = Label("io.docker.compose.service")
+	HASH     = Label("io.docker.compose.config-hash")
+	RECREATE = Label("io.docker.compose.recreate")
```
dnephin: I don't think this label is used anywhere

imikushin: https://github.com/imikushin/libcompose/blob/rebuild/docker/service.go#L216-L217

dnephin: I thought we were going to remove that behaviour?

It's not supported by docker-compose, and it's not actually set anywhere from libcompose, which means it relies on something external to set it?

dnephin: I think a better way to implement it would making the `forceRecreate` param we talk about here: https://github.com/docker/libcompose/pull/89/files#r43012067

imikushin: How would I mark one particular service as "recreate never" in my compose template? 
That is, when I only need one service to persist and assume others (in the same template) are treated by default.

ibuildthecloud: @dnephin RECREATE is set by the user.  I don't think you really confirmed that we should remove the functionality, but I think the discussion has been lost by updates to the PR.  We can't just remove the ability for users to control the recreate behavior through labels.  I really like the feature, actually need the feature, but we will just do it in RancherOS code I guess.  

Regarding https://github.com/docker/libcompose/pull/89/files#r43012067 I don't want to change the approach of passing arguments in this PR.  This code is consistent with the rest of the code base so it doesn't make sense to do it differently.  If we think it should be done a different way then lets address that in a different PR.

ibuildthecloud: @dnephin Sorry there was a terrible typo in that last comment I just fixed.  I should have said "We can just remove the ability"...

dnephin: >  If we think it should be done a different way then lets address that in a different PR.

I think that's fair, I don't want to overload this PR

imikushin: @dnephin, Regarding the RECREATE label: true it's not supported by docker-compose now. But if docker-compose adds similar behaviour, it'll need a similar label. 

@ibuildthecloud and I need this feature for RancherOS, because we're using libcompose to manage system services, and some of them (like console) need to be persistent while the others are recreated on boot. Sure we can re-implement parts of libcompose (or the whole thing, as was the case before @ibuildthecloud factored it out from RancherOS and *created* libcompose), but that doesn't sound like progress.

I'd really like to know what's wrong with this feature (marking services with desired recreate behaviour) that makes it unacceptable here. If someone uses this "recreate" label in their compose template with libcompose, does it hurt docker-compose in any way? Sure they might ask docker-compose maintainers to add support for it in a future version, but is it a bad thing?

Another question: is there an official document for libcompose that states it should be a subset of docker-compose, feature-wise?

dnephin: >  But if docker-compose adds similar behaviour, it'll need a similar label.

It already has similar behaviour. It uses the `--no-recreate` and `--force-recreate` to handle it. With the refactor I've suggested (passing them as arguments to the recreate function, instead of through the context) this would be possible in libcompose as well.

> I'd really like to know what's wrong with this feature (marking services with desired recreate behaviour) that makes it unacceptable here

I think I've alluded to it already, but to re-state:

1. It can lead to impossible situations where you have one service that depends on another, but the service is set to never recreate. When the service is recreated, the service that is set to never recreate breaks because the links or volumes aren't pointing at the correct place anymore. How would you handle that?
2. I think the interface is wrong. It should be handled with parameters. Right now it's a magic label that compose reads but never sets. This is a deviation from all the other labels.
3. It hasn't gone through the standard review process for a new feature. Anything we add to docker-compose has a ticket created and a discussion of the design and implementation. This new feature is being mixed in with a feature parity branch, which is not the right place for it.
4.  It doesn't need to be a compose label, with parameters it could be a rancherOS label, which I think is an argument for keeping it out of compose, at least until there is some more interest in such a feature. 

> Sure we can re-implement parts of libcompose ...

I think with the refactoring I've suggested (to move it to an argument) you won't need to re-implement anything. You'll be able to store the config on the RancherOS side as a label. You can then read the label or config to determine how a conatiner should be recreated.

> is there an official document for libcompose that states it should be a subset of docker-compose, feature-wise?

The README states as the first line "A Go library for Docker Compose. It does everything the command-line tool does". It doesn't say anything about new features really.

I don't think it necessarily has to be a subset of features, but new features should go through the same design review that we have for all projects https://github.com/docker/docker/blob/master/project/REVIEWING.md#workflow

ibuildthecloud: @dnephin 100% agreed, we will rip out the labels.  I think there is a bit of a confusion in that libcompose already supports `--no-recreate` and `--force-recreate` in this PR.  The problem was that it additionally supported the same behavior from labels.  So we will rip out the labels.  We will ensure that libcompose is a strict subset of docker-compose going forward.  The PR was actually created way back and has been dangling for awhile, so I think it wasn't clear how strictly we would conform to docker-compose at that time.  You've made you point loud and clear, so we will always ensure that we don't add new behavior.  Thanks!

docker/container.go
imikushin: Addressed.

docker/container.go
imikushin: Addressed.

docker/service.go
imikushin: Addressed.

project/hash.go
imikushin: Addressed.

cli/command/command.go
imikushin: Will do in a separate PR.

docker/labels.go
imikushin: Addressed.

imikushin:
@dnephin @ibuildthecloud I believe, I've addressed all your comments. 
Now that the controversial "recreate" label has been removed, is this PR ready to be merged?

ibuildthecloud:
@dnephin I think everything is addressed except your comment to pass the convergence strategy as a parameter.  I understand what you are saying there and agree it's better.  Right now though the internals of how services are traversed and what not is actually not correct I believe and the code is pretty dirty.  I'd like to do a separate PR that makes the internals of the code more in line with convergence strategy in docker-compose.  In short, I think passing the flags as arguments can be done in a separate PR.

ibuildthecloud:
LGTM

docker/functions.go
```diff
@@ -34,3 +34,24 @@ func GetContainerByName(client *dockerclient.Client, name string) (*dockerclient
 
 	return &containers[0], nil
 }
+
+// GetContainerByID looks up the hosts containers with the specified Id and
+// returns it, or an error.
+func GetContainerByID(client *dockerclient.Client, id string) (*dockerclient.APIContainers, error) {
+	containers, err := client.ListContainers(dockerclient.ListContainersOptions{All: true, Size: false})
```
vdemeester: I wonder why we get all the container and then filter them out ? `ListContainersOptions` and the API has a way to use filter, and one of them is `id`, so we would let `docker` do the hard work here. Something like :

```go
containers, err := client.ListContainers(dockerclient.ListContainersOptions{
    All: true, 
    Filter: map[string][]string{
        "id":id,
    }, 
    Size: false,
})
```

Then, the only thing is to check the length :wink:

docker/service.go
```diff
@@ -46,15 +59,20 @@ func (s *Service) collectContainers() ([]*Container, error) {
 
 	result := []*Container{}
 
+	if len(containers) == 0 {
```
vdemeester: I'm not sure if it's an optimization or not. If `len(containers)` is equal to `0`, the `for` loop won't do a thing anyway right ? :stuck_out_tongue:

docker/service.go
```diff
@@ -186,15 +188,48 @@ func (s *Service) up(imageName string, create bool) error {
...
+		return err
+	}
+	name := containerInfo.Name[1:]
+
+	logrus.WithFields(logrus.Fields{
+		"outOfSync":     outOfSync,
+		"ForceRecreate": s.context.ForceRecreate,
+		"NoRecreate":    s.context.NoRecreate}).Debug("Going to decide if recreate is needed")
+
+	if s.context.ForceRecreate || outOfSync {
```
vdemeester: We could bring that up (before `containerInfo, err := …`) as we already have all information at this point (except the `name` but see comment below) and this would remove one API call (an inspect done in `c.findInfo()`).

docker/service.go
```diff
@@ -186,15 +188,48 @@ func (s *Service) up(imageName string, create bool) error {
...
+		logrus.Infof("Recreating %s", name)
+		if _, err := c.Recreate(imageName); err != nil {
+			return err
+		}
+	} else if outOfSync {
```
vdemeester: Not sure I get this `if` condition : If `outOfSync` is true, no matter `ForceRecreate`, it got into the first one (and did the `Recreate`), and if `ForceRecreate` is true, it got into the first one too.
So I don't see how this `else if …` is ever gonna be executed. (I might miss something :sweat_smile:)

integration/up_test.go
```diff
@@ -25,6 +26,223 @@ func (s *RunSuite) TestUpNotExistService(c *C) {
...
+
+func mountSet(slice []dockerclient.Mount) map[string]bool {
+	result := map[string]bool{}
+	for _, v := range slice {
+		result[fmt.Sprint(v.Source, ":", v.Destination)] = true
+	}
+	return result
+}
+
+func filter(s map[string]bool, f func(x string) bool) map[string]bool {
```
vdemeester: This method looks useful enough to get into some `utils` package or something (with a unit test :heart_eyes_cat:.

integration/up_test.go
```diff
@@ -25,6 +26,223 @@ func (s *RunSuite) TestUpNotExistService(c *C) {
...
+	p = s.FromText(c, p, "up", "--no-recreate", `
+	hello:
+	  labels:
+	    key: val
+	  image: busybox
+	  stdin_open: true
+	  tty: true
+	`)
+	cn2 := s.GetContainerByName(c, name)
+	c.Assert(cn.ID, Equals, cn2.ID)
```
vdemeester: Shouldn't we also test the labels are not the one in the yaml ?

vdemeester:
Few comments :stuck_out_tongue_closed_eyes:

docker/service.go
imikushin: u'r right, it's just a piece of cruft

docker/service.go
imikushin: Right, it's useless :scissors:

docker/functions.go
imikushin: How come it's not documented in the Docker API? :space_invader: 
But it's there :smiling_imp:

vdemeester: Oh was going to say "yeah i is documented" but.. http://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#list-containers doesn't seem to be up-to-date :sweat_smile:.

docker/service.go
imikushin: Thanks! This one nagged me too. Just a tiny bit: apparently not enough. Fixing.

integration/up_test.go
imikushin: You mean, `filter`? You got it!

imikushin:
@vdemeester Done :wink:

integration/up_test.go
imikushin: Done!

vdemeester:
LGTM 🐸
Thanks @imikushin

imikushin:
Happy to finally see it merged! 
Cheers :beers:

PR #91, opened 2015-10-21
Migrate to candiedyaml from go-yaml

### LINKED ISSUE #79: Switch to non-LGPL yaml parser
https://github.com/go-yaml/yaml is licensed under [LGPL with exceptions](https://github.com/go-yaml/yaml#license).  While legally it is probably okay to run this software many organizations have a strict no GPL of any kind allowed policy for libraries.  I think it is worthwhile investigating alternative yaml libraries.  One possible library is https://github.com/cloudfoundry-incubator/candiedyaml.

imikushin:
Fixes #79.

One gotcha: I had to patch candiedyaml just a tiny bit (see cloudfoundry-incubator/candiedyaml#16) to fix a panic, 
so I'm including my fork instead of cloudfoundry-incubator's. I guess it can take a while before they accept that PR.


dnephin:
I'm a little weary of this change because candiedyaml doesn't seem to be widely used or actively developed. It sounds like my concerns are somewhat justified considering we've already hit a bug with it.

I'd rather wait and see how responsive they are to fixing bugs, instead of having to maintain a separate fork of this library. The only reason to change to a new yaml library is concern over the license, but we haven't had any first-hand reports of this problem (that I know of), so I don't think this change is really urgent.

gondor:
Have you guys looked at github.com/ghodss/yaml ?  I'm not sure if MIT license would conflict but I have used it in man projects with no issues. Just wanted to throw it out there if you haven't seen it.

dnephin:
@gondor I looked into it. I still uses `go-yaml` under the hood, so it doesn't address the licensing issue.

gondor:
@dnephin Ok, good to know thanks!  Side note, noticed your in Toronto.  I'm originally from Canada and went to school at York University in Toronto.  Was a fun time period

ibuildthecloud:
@dnephin I'm not too concerned about this library.  The approach taken is basically the same as go-yaml in that its a transpilation of libyaml from c to go.  So the core should be fairy good.

imikushin:
My PR to candiedyaml (cloudfoundry-incubator/candiedyaml#16) got merged, so I'm updating this PR with the latest upstream candiedyaml code instead of my own fork. 

/cc @ibuildthecloud @dnephin @vdemeester

ibuildthecloud:
LGTM

ibuildthecloud:
ping @vdemeester

ibuildthecloud:
@imikushin Can you ~~rebase~~ recreate on master :smile:

imikushin:
Redone. :smile:

vdemeester:
LGTM 🐶

I'm not really found of the `SliceOrMap` and other struct in `type_yaml.go` but it's not worse so :stuck_out_tongue:.

PR #92, opened 2015-10-21
Fix TLS support

dhiltgen:
This wires up creating the TLS client based on the fsouza API.

Signed-off-by: Daniel Hiltgen <daniel.hiltgen@docker.com>

### DIFF
--- a/docker/client.go
+++ b/docker/client.go
@@ -1,7 +1,6 @@
 package docker
 
 import (
-	"crypto/tls"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
@dhiltgen I feel a bit more confortable with using tlsconfig pkg to build the TLSConfig as it's the same package used by the docker client.  Is TLS currently broken with the fsouza client?  Is there anyway we can continue to use `pkg/tlsconfig` and use the fsouza client?

dhiltgen:
This was the first time I've used libcompose, and it was DOA for TLS configured systems for me.  If it works for others, maybe the problem is on my side, but this change fixed it for me.

ibuildthecloud:
@dhiltgen I don't doubt that TLS broke when we switched clients.  @imikushin Did you by chance test TLS?  Testing against docker-machine is usually the easiest for me since it sets up TLS for you.

imikushin:
I confirm this PR fixes TLS support. The only problem is golint is pissed off. 
I'll rebase this PR and try to calm golint.

vdemeester:
Carried by #106, closing this one. Thanks @dhiltgen

PR #93, opened 2015-10-21
Don't use memswap_limit for test

vdemeester:
As it could be disabled on the machine that builds and thus give an test
failure for a wrong reason.

See https://jenkins.dockerproject.org/job/Libcompose%20Master/71/console


ibuildthecloud:
LGTM

PR #105, opened 2015-11-09
Bump to docker 1.10-dev and dependencies

### LINKED ISSUE #32: libcompose should be go gettable/buildable/testable

vdemeester:
This is more a *try* (or a *"it would look like this"* kinda thing) than a real thing because this feels almost too early to bump a docker version but as we still don't have versions, so it feel safer (will probably just have to bump it on the way later too).

Anyway, what this does is bumping to current master (my dream is to bump it on a tag but… it's not gonna happen until `1.10`) to remove the `autogen/dockerversion` thingy and make it *go gettable/buildable/testable*, and thus would fix #32.

- go gettable/…
- No more use of `go generate`,
- Keep the source up-to-date and update a little tiny bit the code base to compile with it :wink:.

@ibuildthecloud @docker/libcompose-maintainers WDYT ?


ibuildthecloud:
I like the idea, and I'm fine with pulling this in before 1.10 is released.  It's almost safer to bump now as it's closer to 1.9 as it is right now.

vdemeester:
@ibuildthecloud cool :smile_cat: Then I have to see why the integration tests are failing and fix it :stuck_out_tongue_closed_eyes:.

vdemeester:
Updated with some fixes. I splited the commit into two, one for the vendoring, the other one for the fix.
Build should be green :wink:.

gondor:
Great idea @vdemeester ! Very happy with this :) as it will be go gettable it appears

vdemeester:
ping @ibuildthecloud @dnephin

dnephin:
Changes in the second commit seem fine. 
I think it's ok to use a pre-release version of Docker.
Can't really review the `vendor/` change

LGTM

imikushin:
@vdemeester Does this work with docker-1.9.1? If it does, maybe vendor it?

vdemeester:
@imikushin I could update the "hash" but the `dockerversion` thingy was not cherry-picked on `1.9.1` so the whole point of this PR wouldn't be met with `1.9.1` :stuck_out_tongue_closed_eyes:

imikushin:
@vdemeester to get rid of `go generate`? okay :+1:

gondor:
@vdemeester Do you have any timeline on when this will be merged?

ibuildthecloud:
LGTM

PR #110, opened 2015-11-25
small yaml marshaling cleanup

imikushin:
small yaml marshaling cleanup

### DIFF
--- a/project/types_yaml_test.go
+++ b/project/types_yaml_test.go
@@ -120,6 +120,11 @@ type StructSliceorMap struct {
 	Bars []string   `yaml:"bars"`
 }
 
+type StructCommand struct {
+	Entrypoint Command `yaml:"entrypoint,flow,omitempty"`
+	Command    Command `yaml:"command,flow,omitempty"`
+}
+
 func TestSliceOrMapYaml(t *testing.T) {
 	str := `{foos: [bar=baz, far=faz]}`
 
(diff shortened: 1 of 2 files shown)


dnephin:
I'm not sure I fully understand this change, can you tell me more about it?

Is s.parts never going to be nil?

imikushin:
The code is just redundant. It was necessary with go-yaml, but isn't needed now.

project/types_yaml_test.go
```diff
@@ -192,3 +197,24 @@ func TestMaporsliceYaml(t *testing.T) {
...
+	bytes, err := yaml.Marshal(s)
+	assert.Nil(t, err)
+
+	s2 := &StructCommand{}
+	err = yaml.Unmarshal(bytes, s2)
+
+	assert.Nil(t, err)
+	assert.Equal(t, []string{"bash"}, s.Command.Slice())
+	assert.Nil(t, s.Entrypoint.Slice())
+}
```
dnephin: I don't think this test actually tests the change in behaviour because the command is not empty.

dnephin:
But why is it redundant now?  Does the new yaml parser default to the empty string slice instead of nil?

The test suggests that it is still using nil for entrypoint, no?

imikushin:
The code is redundant because it doesn't change behaviour, right? 

I've written the just to make sure the unmarshal/marshal result is the same before and after the change in test_yaml.go 

Maybe it's useless.

dnephin:
I guess what I'm wondering is: does it change the behaviour, but we just aren't testing the behaviour it changed? 

My understanding of the old code is that it was replacing a `nil` with an empty string slice, presumably so that code later on doesn't have to check for a nil.  The test confirms that marshal/unmarshal works, but not that the rest of the code in libcompose is going to work with a nil in these fields.

ibuildthecloud:
LGTM

@dnephin for all practical purposes in go `nil` and `[]string{}` are practically the same.  About the only diffference is that `x == nil` will be true for one and false for the other.  But the following works

```
var empty []string = nil
len(empty) // 0
append(empty, "foo") // ["foo"]
```

dnephin:
Ok, I think that's what I was missing. I was forgetting that golang auto-initializes these as necessary. 

Thanks
LGTM

vdemeester:
LGTM :wink:

PR #111, opened 2015-11-25
Do not transplant ServiceConfig inner slices and maps during Convert().

imikushin:
Do not transplant ServiceConfig inner slices and maps during Convert().

Copy them instead. Otherwise, if we change e.g. labels on the resulting container config,
we'll have these changes injected back into the ServiceConfig being unaware of that.

### DIFF
--- a/utils/util.go
+++ b/utils/util.go
@@ -70,6 +70,28 @@ func Convert(src, target interface{}) error {
 	return err
 }
 
+// CopySlice creates an exact copy of the provided string slice
+func CopySlice(s []string) []string {
+	if s == nil {
+		return nil
+	}
+	r := make([]string, len(s))
+	copy(r, s)
+	return r
+}
+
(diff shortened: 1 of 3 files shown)


utils/util.go
```diff
@@ -70,6 +70,30 @@ func Convert(src, target interface{}) error {
...
+func CopySlice(s []string) []string {
+	if s == nil {
+		return nil
+	}
+	r := []string{}
+	for _, v := range s {
+		r = append(r, v)
+	}
+	return r
+}
```
dnephin: Is there a reason https://golang.org/pkg/builtin/#copy won't work here?

imikushin: Should work, I guess.

imikushin: Updated.

dnephin:
LGTM

vdemeester:
LGTM 🐰

PR #116, opened 2015-12-03
Multiple -f

### LINKED ISSUE #98: Support multiple -f
https://github.com/docker/compose/blob/129092b7/docs/reference/docker-compose.md

joshwget:
Fixes #98. This changes some properties of `Context` (`string ComposeFile` to `[]string ComposeFiles` and `[]byte ComposeBytes` to `[][]byte ComposeBytes`) as well as `Project` (`string File` to `[]string Files`).

### DIFF
--- a/cli/command/command.go
+++ b/cli/command/command.go
@@ -1,6 +1,8 @@
 package command
 
 import (
+	"os"
+
 	"github.com/codegangsta/cli"
 	"github.com/docker/libcompose/cli/app"
 	"github.com/docker/libcompose/project"
@@ -229,7 +231,11 @@ func CommonFlags() []cli.Flag {
 // Populate updates the specified project context based on command line arguments and subcommands.
 func Populate(context *project.Context, c *cli.Context) {
 	if len(c.GlobalStringSlice("file")) == 0 {
-		context.ComposeFiles = []string{"docker-compose.yml"}
+		if _, err := os.Stat("docker-compose.override.yml"); err == nil {
+			context.ComposeFiles = []string{"docker-compose.yml", "docker-compose.override.yml"}
+		} else {
+			context.ComposeFiles = []string{"docker-compose.yml"}
+		}
 	} else {
 		context.ComposeFiles = c.GlobalStringSlice("file")
 	}
--- a/cli/command/command.go
+++ b/cli/command/command.go
@@ -213,10 +213,10 @@ func CommonFlags() []cli.Flag {
 		cli.BoolFlag{
 			Name: "verbose,debug",
 		},
-		cli.StringFlag{
+		cli.StringSliceFlag{
 			Name:   "file,f",
-			Usage:  "Specify an alternate compose file (default: docker-compose.yml)",
-			Value:  "docker-compose.yml",
+			Usage:  "Specify one or more alternate compose files (default: docker-compose.yml)",
+			Value:  &cli.StringSlice{},
 			EnvVar: "COMPOSE_FILE",
 		},
 		cli.StringFlag{
@@ -228,7 +228,12 @@ func CommonFlags() []cli.Flag {
 
 // Populate updates the specified project context based on command line arguments and subcommands.
 func Populate(context *project.Context, c *cli.Context) {
-	context.ComposeFile = c.GlobalString("file")
+	if len(c.GlobalStringSlice("file")) == 0 {
+		context.ComposeFiles = []string{"docker-compose.yml"}
+	} else {
+		context.ComposeFiles = c.GlobalStringSlice("file")
+	}
+
 	context.ProjectName = c.GlobalString("project-name")
 
 	if c.Command.Name == "logs" {
--- a/project/context.go
+++ b/project/context.go
@@ -24,8 +24,8 @@ type Context struct {
 	ForceRecreate       bool
 	NoRecreate          bool
 	Signal              int
-	ComposeFile         string
-	ComposeBytes        []byte
+	ComposeFiles        []string
+	ComposeBytes        [][]byte
 	ProjectName         string
 	isOpen              bool
 	ServiceFactory      ServiceFactory
@@ -36,32 +36,35 @@ type Context struct {
 	Project             *Project
 }
 
-func (c *Context) readComposeFile() error {
+func (c *Context) readComposeFiles() error {
 	if c.ComposeBytes != nil {
 		return nil
 	}
 
-	logrus.Debugf("Opening compose file: %s", c.ComposeFile)
+	logrus.Debugf("Opening compose files: %s", strings.Join(c.ComposeFiles, ","))
 
-	if c.ComposeFile == "-" {
+	if len(c.ComposeFiles) == 1 && c.ComposeFiles[0] == "-" {
 		composeBytes, err := ioutil.ReadAll(os.Stdin)
 		if err != nil {
 			logrus.Errorf("Failed to read compose file from stdin: %v", err)
 			return err
 		}
-		c.ComposeBytes = composeBytes
-	} else if c.ComposeFile != "" {
-		if composeBytes, err := ioutil.ReadFile(c.ComposeFile); os.IsNotExist(err) {
-			if c.IgnoreMissingConfig {
-				return nil
+		c.ComposeBytes = [][]byte{composeBytes}
+	} else {
+		for _, composeFile := range c.ComposeFiles {
+			if composeFile != "" {
+				if composeBytes, err := ioutil.ReadFile(composeFile); os.IsNotExist(err) {
+					if !c.IgnoreMissingConfig {
+						logrus.Errorf("Failed to find %s", composeFile)
+						return err
+					}
+				} else if err != nil {
+					logrus.Errorf("Failed to open %s", composeFile)
+					return err
+				} else {
+					c.ComposeBytes = append(c.ComposeBytes, composeBytes)
+				}
 			}
-			logrus.Errorf("Failed to find %s", c.ComposeFile)
-			return err
-		} else if err != nil {
-			logrus.Errorf("Failed to open %s", c.ComposeFile)
-			return err
-		} else {
-			c.ComposeBytes = composeBytes
 		}
 	}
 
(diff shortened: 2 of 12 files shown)


dnephin:
I don't think this handles merge the same way as compose. For one, it needs to merge `extends`, `links` and `volumes_from`, which is not done by `extends.

joshwget:
@dnephin If `links` and `volume_from` are merged with multiple files the same way as when using `extends`, I think this will support those. It definitely doesn't merge `extends` though. I'll address this.

ibuildthecloud:
@joshwget I like the approach, took me a bit to notice you are storing `[][]byte`.

@dnephin Is the merging behavior of extends documented?  From what I understand of this PR is if you do `-f a.yml -f b.yml` it will evaluate the extends in a.yml and b.yml before merging.  That seems like the correct behavior to me, but I'm not sure what docker-compose is actually doing.

dnephin:
That sounds right

joshwget:
@dnephin I think that addresses all of your original comments then, but please let me know if I missed something!

vdemeester:
@joshwget According to the [documentaiton](https://github.com/docker/compose/blob/master/docs/extends.md#multiple-compose-files) compose looks by default for `docker-compose.yml` and an optionnal `docker-compose.override.yml`.

joshwget:
Thanks @vdemeester, should be addressed now!

vdemeester:
Hum… doesn't seem to work as expected, I'm using the `test/fixtures/multiple-composefiles` files from the `docker/compose` project :

```bash
# With docker-compose
$ docker-compose -f docker-compose.yml -f compose2.yml up -d
Creating multiplecomposefiles_simple_1
Creating multiplecomposefiles_yetanother_1
Creating multiplecomposefiles_another_1
$ docker-compose -f docker-compose.yml -f compose2.yml ps
Name                  Command   State   Ports 
-----------------------------------------------------------
multiplecomposefiles_another_1      top       Up            
multiplecomposefiles_simple_1       top       Up            
multiplecomposefiles_yetanother_1   top       Up  

# With libcompose-cli on this PR
$ libcompose-cli -f docker-compose.yml -f compose2.yml up -d
INFO[0000] Project [multiple-composefiles]: Starting project  
INFO[0000] [0/1] [yetanother]: Starting                 
INFO[0000] [1/1] [yetanother]: Started  
$ libcompose-cli -f docker-compose.yml -f compose2.yml ps 
Name                                Command  State          Ports
multiple-composefiles_yetanother_1  top      Up 29 seconds  
```

It only picked that latest one and did no merge :sweat_smile:.

joshwget:
@vdemeester When I try reproducing your example I get results similar to Compose. Would you mind checking this again? :smile:

vdemeester:
@joshwget you are right, I might have messed up, it's working as expected :stuck_out_tongue_closed_eyes: ; with `docker-compose.override.conf` and such.

cli/command/command.go
```diff
@@ -228,7 +230,16 @@ func CommonFlags() []cli.Flag {
 
 // Populate updates the specified project context based on command line arguments and subcommands.
 func Populate(context *project.Context, c *cli.Context) {
-	context.ComposeFile = c.GlobalString("file")
+	if len(c.GlobalStringSlice("file")) == 0 {
```
vdemeester: I would have written like this (because I don't like `else`'s :stuck_out_tongue_closed_eyes:), but not sure if it's better or not...

```go
	context.ComposeFiles = c.GlobalStringSlice("file")

	if len(context.ComposeFiles) == 0 {
		context.ComposeFiles = []string{"docker-compose.yml"}
		if _, err := os.Stat("docker-compose.override.yml"); err == nil {
			context.ComposeFiles = append(context.ComposeFiles, "docker-compose.override.yml")
		}
	}
```

project/context.go
```diff
@@ -36,32 +36,35 @@ type Context struct {
...
 		}
-		c.ComposeBytes = composeBytes
-	} else if c.ComposeFile != "" {
-		if composeBytes, err := ioutil.ReadFile(c.ComposeFile); os.IsNotExist(err) {
-			if c.IgnoreMissingConfig {
-				return nil
+		c.ComposeBytes = [][]byte{composeBytes}
+	} else {
+		for _, composeFile := range c.ComposeFiles {
+			if composeFile != "" {
```
vdemeester: I'm not sure we should do this check, `docker-compose` will fail if we manage to specify an empty compose file :

```bash
$ libcompose-cli -f ../ports-composefile/docker-compose.yml -f "" up -d
INFO[0000] Project [ports-composefile]: Starting project  
INFO[0000] [0/1] [simple]: Starting                     
INFO[0000] [1/1] [simple]: Started                      

$ docker-compose -f ../ports-composefile/docker-compose.yml -f "" up -d
ERROR: .IOError: [Errno 21] Is a directory: u'./'
```

project/context.go
```diff
@@ -36,32 +36,35 @@ type Context struct {
 	Project             *Project
 }
 
-func (c *Context) readComposeFile() error {
+func (c *Context) readComposeFiles() error {
```
vdemeester: This method feels a little bit complicated to me (too much complexity / indentation).

```go
func (c *Context) readComposeFiles() error {
	if c.ComposeBytes != nil {
		return nil
	}

	logrus.Debugf("Opening compose files: %s", strings.Join(c.ComposeFiles, ","))

	// Handle STDIN (`-f -`)
	if len(c.ComposeFiles) == 1 && c.ComposeFiles[0] == "-" {
		composeBytes, err := ioutil.ReadAll(os.Stdin)
		if err != nil {
			logrus.Errorf("Failed to read compose file from stdin: %v", err)
			return err
		}
		c.ComposeBytes = [][]byte{composeBytes}
		return nil
	}

	for _, composeFile := range c.ComposeFiles {
		composeBytes, err := ioutil.ReadFile(composeFile)
		if err != nil {
			if os.IsNotExist(err) && !c.IgnoreMissingConfig {
				logrus.Errorf("Failed to find %s", composeFile)
				return err
			}
			logrus.Errorf("Failed to open %s", composeFile)
			return err
		}
		c.ComposeBytes = append(c.ComposeBytes, composeBytes)
	}

	return nil
}
```

or even :

```go
func (c *Context) readComposeFiles() error {
	if c.ComposeBytes != nil {
		return nil
	}

	logrus.Debugf("Opening compose files: %s", strings.Join(c.ComposeFiles, ","))

	// Handle STDIN (`-f -`)
	if len(c.ComposeFiles) == 1 && c.ComposeFiles[0] == "-" {
		composeBytes, err := readComposeFile(os.Stdin)
		if err != nil {
			return err
		}
		c.ComposeBytes = [][]byte{composeBytes}
		return nil
	}

	for _, composeFile := range c.ComposeFiles {
		composeFileF, err := os.Open(composeFile)
		if os.IsNotExist(err) && !c.IgnoreMissingConfig {
			logrus.Errorf("Failed to find %s", composeFile)
			return err
		}
		composeBytes, err := readComposeFile(composeFileF)
		if err != nil {
			return err
		}
		c.ComposeBytes = append(c.ComposeBytes, composeBytes)
	}

	return nil
}

func readComposeFile(reader io.Reader) ([]byte, error) {
	composeBytes, err := ioutil.ReadAll(reader)
	if err != nil {
		logrus.Errorf("Failed to read compose file : %v", err)
		return nil, err
	}
	return composeBytes, nil
}
```

After looking at both, I prefer the first one :smiley_cat:.

project/context.go
```diff
@@ -96,9 +99,14 @@ func (c *Context) lookupProjectName() (string, error) {
 		return envProject, nil
 	}
 
-	f, err := filepath.Abs(c.ComposeFile)
+	file := ""
```
vdemeester: I would use `file := "."` to show that it takes the current directory by default — it's a nit, really :stuck_out_tongue_closed_eyes: .

project/merge.go
```diff
@@ -44,13 +44,23 @@ func mergeProject(p *Project, bytes []byte) (map[string]*ServiceConfig, error) {
...
-		data, err := parse(p.context.ConfigLookup, p.context.EnvironmentLookup, p.File, data, datas)
+		data, err := parse(p.context.ConfigLookup, p.context.EnvironmentLookup, file, data, datas)
+
 		if err != nil {
 			logrus.Errorf("Failed to parse service %s: %v", name, err)
 			return nil, err
 		}
 
-		datas[name] = data
+		if _, ok := p.Configs[name]; ok {
```
vdemeester:
```go

		if _, ok := p.Configs[name]; ok {
			var rawExistingService rawService
			if err := utils.Convert(p.Configs[name], &rawExistingService); err != nil {
				return nil, err
			}

			data = mergeConfig(rawExistingService, data)
		}
		
		datas[name] = data
```

project/project.go
```diff
@@ -63,14 +63,24 @@ func (p *Project) Parse() error {
 
 	p.Name = p.context.ProjectName
 
-	if p.context.ComposeFile == "-" {
-		p.File = "."
+	if len(p.context.ComposeFiles) == 1 && p.context.ComposeFiles[0] == "-" {
```
vdemeester:
```go
	p.Files = p.context.ComposeFiles

	if len(p.Files) == 1 && p.Files[0] == "-" {
		p.Files = []string{"."}
	}
```

project/project.go
```diff
@@ -63,14 +63,24 @@ func (p *Project) Parse() error {
...
+		p.Files = []string{"."}
 	} else {
-		p.File = p.context.ComposeFile
+		p.Files = p.context.ComposeFiles
 	}
 
 	if p.context.ComposeBytes != nil {
-		return p.Load(p.context.ComposeBytes)
+		for i, composeBytes := range p.context.ComposeBytes {
+			if i < len(p.context.ComposeFiles) {
```
vdemeester: What about :

```go

			file := ""
			if i < len(p.context.ComposeFiles) {
				file = p.Files[i]
			}
			if err := p.load(file, composeBytes); err != nil {
				return err
			}
```

But then, I would refactor the `Load` method to take a file *string* and not create `load` (if not used elsewhere).

vdemeester:
@joshwget So far, so good :wink:. Few commands on code but it almost work as expected, that's cool :+1: :wink:.

I would add more integration though :
- Using the default `docker-compose.yml` and `docker-compose.override.yml`
- Pointing to one or several files without specifying the project name and verifying that the project name is named based on the first folder.

vdemeester:
Carrying into #140, thanks for the hard work @joshwget

PR #121, opened 2015-12-17
Acceptance tests

shin-:
This PR adds the ability to run the [acceptance test suite](https://github.com/docker/compose/tree/master/tests/acceptance) from docker/compose against the libcompose CLI. It is configured to run with the `test` make target (or individually with `test-acceptance`).

cc @dnephin

### DIFF
--- a/script/test-acceptance
+++ b/script/test-acceptance
@@ -1,0 +1,22 @@
+#!/bin/bash
+set -e
+
+export SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+export DEST=.
+
+bundle .integration-daemon-start
+trap "bundle .integration-daemon-stop" EXIT
+
+rm -rf venv
+virtualenv venv
+git clone https://github.com/docker/compose.git venv/compose
+venv/bin/pip install \
+    -r venv/compose/requirements.txt \
+    -r venv/compose/requirements-dev.txt
+
+cp bundles/libcompose-cli_linux-amd64 venv/bin/docker-compose
+. venv/bin/activate
+
+docker-compose --version
+cd venv/compose
+py.test -vs --tb=short tests/acceptance
--- a/script/test-acceptance
+++ b/script/test-acceptance
@@ -5,7 +5,6 @@ export SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 export DEST=.
 
 bundle .integration-daemon-start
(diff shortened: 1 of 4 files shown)


script/test-acceptance
```diff
@@ -0,0 +1,34 @@
...
+set -e
+
+export SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+export DEST=.
+
+bundle .integration-daemon-start
+
+apt-get install -y python-setuptools
+easy_install pip
+pip install virtualenv
```
vdemeester: @shin- I would definitely put these steps in the `Dockerfile` and not there

- I don't want it to run each time I run `script/test-acceptance`
- I want to be able to run this script (outside the docker container, why not :stuck_out_tongue_closed_eyes:) on my Archlinux for example (but could be on a fedora, …) where, I don't have `apt-get`

These (`setuptools`, `pip` and `virtualenv`) are requirements to execute this test-acceptance, but they definitely shouldn't be in the script.

vdemeester:
@shin- Very cool :raised_hands: I Added a remark on the script

How do we handle the "missing" part (stuff that do not work or not yet implemented) for now ? There is 2 options I see : 
- Waiting for them to get fixed (and rebasing / using this PR/branch to validate this) and then merge
- Merge this one but make the build success if the acceptance test fails ; and when it's all *conform*, we can re-enable *strict* checking on the acceptance test.

@docker/libcompose-maintainers wdyt ?

PS : I'm preparing an issue to keep track of the missing part libcompose has with docker-compose 1.5 so we can focus on working on these :wink:.

script/test-acceptance
dnephin: sounds right to me

aanand:
I think I'm leaning toward making it a non-critical part of the test suite, so that it doesn't hold up work on anything else. It seems about the right level of annoying to have it produce a lot of failing output, but not actually break the build.

vdemeester:
I didn't answer my own question, I'm also for a non-critical part for now :)

shin-:
Updated addressing (hopefully) everyone's comments. Thank you!

dnephin:
LGTM!

vdemeester:
@shin- could you squash your 4 last commits ? and it will definitely LGTM for me :stuck_out_tongue_closed_eyes:

shin-:
Squashed :+1:

aanand:
Nice one.

PR #122, opened 2015-12-18
Unmarshal empty string slices as nil

imikushin:
Unmarshal empty string slices as nil. 

The problem is: if we have an empty slice as a command or entrypoint (e.g. in marshaled service config) 
docker API yells we're passing an empty command or entrypoint. But it doesn't mind having nils. 

Also, bump test dependencies for nicer unit test errors. Bump Go as well.


ibuildthecloud:
LGTM

aanand:
Thanks. Could you separate the version bumps out into another commit?

vdemeester:
LGTM 🐲 but I do agree with @aanand

imikushin:
Here you are: two separate nice little commits :)

vdemeester:
:heart_eyes_cat:

PR #102, opened 2015-11-06
Two improvements for command help

bohai:
Implement for  Issue #101

PR #126, opened 2015-12-25
Carry #102 : improvements for command help

### LINKED ISSUE #101: Two improvements about libcompose command help
##### No.1   
The command usage  is not displayed for the command which has no options.
Maybe this is a bug.
##### No.2 
Many commands of libcompose support specifying some services as the command args.
But this is not displayed in the command help .

##### Example
Current display:  
<pre><code>
root@SZX1000041895:~/sleep# libcompose-cli help create
Usage: libcompose-cli create
</code></pre>

Should be:
<pre><code>
root@SZX1000041895:~/sleep# ./libcompose-cli help create
Usage: libcompose-cli create [SERVICE...]

Create all services but do not start
</code></pre>

vdemeester:
Closes #101, Carrying #102 and makes help on commands (description, usage and such) more close to what `docker-compose` does 🐰. It is not mandatory but :stuck_out_tongue_closed_eyes:.

It looks like that :

```bash
$ .libcompose-cli help up
Builds, (re)creates, starts, and attaches to containers for a service.

Unless they are already running, this command also starts any linked services.

The `libcompose-cli up` command aggregates the output of each container. When
the command exits, all containers are stopped. Running `libcompose-cli up -d`
starts the containers in the background and leaves them running.

If there are existing containers for a service, and the service's configuration
or image was changed after the container's creation, `libcompose-cli up` picks
up the changes by stopping and recreating the containers (preserving mounted
volumes). To prevent Compose from picking up changes, use the `--no-recreate`
flag.

If you want to force Compose to stop and recreate all containers, use the
`--force-recreate` flag.

Usage: libcompose-cli_linux-amd64 up [options] [SERVICE...]

Options:
   -d                   Do not block and log
   --no-recreate        If containers already exist, don't recreate them. Incompatible with --force-recreate.
   --force-recreate     Recreate containers even if their configuration and image haven't changed. Incompatible with --no-recreate.

$ libcompose-cli help pull
Pulls images for services.

Usage: libcompose-cli_linux-amd64 pull [SERVICE...]
```

- [ ] Find a better way for description
- [ ] Re-order commands (alphabetically, like `docker-compose`)

At some point we should remove the `WARN` and make it clear elsewhere (like on the main help/usage command), but that's another story/PR :wink:.

/cc @dnephin @aanand @ibuildthecloud 

🐸

### DIFF
--- a/cli/command/command.go
+++ b/cli/command/command.go
@@ -9,17 +9,22 @@ import (
 // CreateCommand defines the libcompose create subcommand.
 func CreateCommand(factory app.ProjectFactory) cli.Command {
 	return cli.Command{
-		Name:   "create",
-		Usage:  "Create all services but do not start",
-		Action: app.WithProject(factory, app.ProjectCreate),
+		Name:      "create",
+		Usage:     "Create all services but do not start",
+		ArgsUsage: "[SERVICE...]",
+		Action:    app.WithProject(factory, app.ProjectCreate),
 	}
 }
 
 // BuildCommand defines the libcompose build subcommand.
 func BuildCommand(factory app.ProjectFactory) cli.Command {
 	return cli.Command{
-		Name:   "build",
-		Usage:  "Build or rebuild services.",
+		Name:      "build",
+		Usage:     "Build or rebuild services.",
+		ArgsUsage: "[SERVICE...]",
+		Description: "Services are built once and then tagged as `project_service`,\n" +
+			"e.g. `composetest_db`. If you change a service's `Dockerfile` or the\n" +
+			"contents of its build directory, you can run `libcompose-cli build` to rebuild it.",
 		Action: app.WithProject(factory, app.ProjectBuild),
 		Flags: []cli.Flag{
 			cli.BoolFlag{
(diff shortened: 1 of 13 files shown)


cli/command/command.go
```diff
@@ -69,8 +76,20 @@ func PortCommand(factory app.ProjectFactory) cli.Command {
 // UpCommand defines the libcompose up subcommand.
 func UpCommand(factory app.ProjectFactory) cli.Command {
 	return cli.Command{
-		Name:   "up",
-		Usage:  "Bring all services up",
+		Name:      "up",
+		Usage:     "Builds, (re)creates, starts, and attaches to containers for a service.",
+		ArgsUsage: "[SERVICE...]",
+		Description: "Unless they are already running, this command also starts any linked services.\n\n" +
+			"The `libcompose-cli up` command aggregates the output of each container. When\n" +
```
thaJeztah: It's unfortunate we need the back-tics in the string; having to do all the concatenation and `\n\n` for new lines looks tedious

vdemeester: Yeah :sweat: I'm looking around to see a better way to do that (and hopefully, I'll find one).

thaJeztah: Yup, also thinking what a solution would be. Perhaps a separate TOML file containing the strings/descriptions? idk

vdemeester: @thaJeztah I wonder, I could also use *backticks* to not have to escape the carriage-return but.. I had trouble with them for [these](https://github.com/docker/libcompose/pull/126/files#diff-3145698b159fb0a026984b4a017a8b24R25) (backticks inside backticks…).

thaJeztah: Thinking if GoDoc can be used somehow, I know @calavera did some experimenting in this area in this PR https://github.com/docker/docker/pull/16125. However (obviously) that's used to generate external documentation, not for usage in the output of the software itself

dnephin: spf13/cobra has a good way of dealing with this I think: https://github.com/spf13/cobra#manually-create-rootcmd

thaJeztah: Looks like that also uses back-tics for quotes. Still might be worth looking into; something that's easy to maintain and reusable

vdemeester:
I'm closing this for now, will handle this when migration to `spf13/cobra` :wink:

PR #130, opened 2015-12-26
Fixing rm command and API…

### LINKED ISSUE #129: The rm command always require `--force`
The `rm` only works with specifying the `--force` flag. This does not work as expected and as `docker-compose` does.

<del>The `--force` works as expected when the container are still running.</del> Even the force doesn't work as expected, `docker-compose rm --force` just don't ask for confirmation, but it will not remove running containers at all.

## Steps to reproduce

Let's use the following `docker-compose.yml` :

```yaml
web:
    image: nginx
```

### Using `libcompose-cli`

```bash
$ libcompose-cli up -d
# […]
$ libcompose-cli stop
# […]
$ libcompose-cli rm
# […]
FATA[0000] Will not remove all services without --force 
```

### Using `docker-compose`

```bash
$ docker-compose up -d
Creating compose_web_1
$ docker-compose stop
Stopping compose_web_1 ... done
$ docker-compose rm
Going to remove compose_web_1
Are you sure? [yN] y
Removing compose_web_1 ... done
```

vdemeester:
… to behave more like docker-compose. Fixes #129 🐝.

```bash
$ libcompose-cli ps
# […]
Name           Command                         State         Ports
compose_db_1   /docker-entrypoint.sh postgres  Up 8 seconds  5432/tcp
compose_web_1  nginx -g 'daemon off;'          Up 8 seconds  443/tcp, 80/tcp
$ libcompose-cli rm
# […]                      
No stopped containers
$ libcompose-cli stop db
# […]
$ libcompose-cli rm
# […]
Going to remove compose_web_1, compose_db_1
Are you sure? [yN]
```

As `docker-compose` it will try to remove only containers that are stopped. And the `--force` flag is used to skip the confirmation only 🐞.

```bash
$ libcompose-cli stop db
# […]
$ libcompose-cli rm --force
# […]                  
Going to remove compose_db_1
$ libcompose-cli ps
# […]
Name           Command                 State          Ports
compose_web_1  nginx -g 'daemon off;'  Up 54 seconds  443/tcp, 80/tcp
```

It's not really done yet :
- [X] <del>Probably should add some log/information of each container removed (like `docker-compose` does).</del> This should be handled widely on the cli I think.
- [X] Should add more integration tests (making sure we don't remove non-stopped containers, etc…).

🐸

### DIFF
--- a/docker/container.go
+++ b/docker/container.go
@@ -210,14 +210,26 @@ func (c *Container) Delete() error {
 		return err
 	}
 
-	if info.State.Running {
-		err := c.client.StopContainer(container.ID, c.service.context.Timeout)
-		if err != nil {
-			return err
-		}
+	if !info.State.Running {
+		return c.client.RemoveContainer(dockerclient.RemoveContainerOptions{ID: container.ID, RemoveVolumes: c.service.context.Volume})
 	}
 
-	return c.client.RemoveContainer(dockerclient.RemoveContainerOptions{ID: container.ID, Force: true, RemoveVolumes: c.service.context.Volume})
+	return nil
+}
+
+// IsRunning returns the running state of the container.
+func (c *Container) IsRunning() (bool, error) {
+	container, err := c.findExisting()
+	if err != nil || container == nil {
+		return false, err
+	}
+
+	info, err := c.client.InspectContainer(container.ID)
+	if err != nil {
+		return false, err
+	}
+
+	return info.State.Running, nil
 }
 
 // Up creates and start the container based on the image name and send an event
--- a/project/project.go
+++ b/project/project.go
@@ -226,6 +226,39 @@ func (p *Project) Pull(services ...string) error {
 	}), nil)
 }
 
+// ListStoppedContainers lists the stopped containers for the specified services.
+func (p *Project) ListStoppedContainers(services ...string) ([]string, error) {
+	stoppedContainers := []string{}
+	err := p.forEach(services, wrapperAction(func(wrapper *serviceWrapper, wrappers map[string]*serviceWrapper) {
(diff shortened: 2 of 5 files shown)


docker/container.go
```diff
@@ -190,14 +190,26 @@ func (c *Container) Delete() error {
 		return err
 	}
 
-	if info.State.Running {
-		err := c.client.StopContainer(container.ID, c.service.context.Timeout)
-		if err != nil {
-			return err
-		}
+	if !info.State.Running {
+		return c.client.RemoveContainer(dockerclient.RemoveContainerOptions{ID: container.ID, Force: true, RemoveVolumes: c.service.context.Volume})
```
dnephin: I don't think we want `Force: true`. We're already checking if it's stopped, so we shouldn't need to force it.

project/project.go
```diff
@@ -226,6 +226,35 @@ func (p *Project) Pull(services ...string) error {
...
+				return innerErr
+			}
+
+			for _, container := range containers {
+				running, innerErr := container.IsRunning()
+				if innerErr != nil {
+					log.Error(innerErr)
+				}
+				if !running {
+					stoppedContainers = append(stoppedContainers, container.Name())
```
dnephin: Shouldn't this be container id ?

docker/container.go
vdemeester: Oh you're right :stuck_out_tongue_closed_eyes:.

project/project.go
vdemeester: @dnephin it could be the `ID` but both names and ids are unique so I feels it would do the same (but yeah using `ID` is ok, I'll update).

vdemeester:
@dnephin Updated, rebased and added an integration tests.

dnephin:
LGTM

PR #133, opened 2015-12-29
Implement run command

vdemeester:
Carrying #45, add a basic `run` command (with no flags for now). 🐨

```bash
# libcompose run <service> <command>
$ ~docker/libcompose/libcompose-cli run web ls
bin  boot  dev  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
$ ~docker/libcompose/libcompose-cli run web /bin/sh
# ls
bin  boot  dev  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
# echo foo > /bar    
# cat bar
foo
```

This is the first step to have a run command that is acting the same as docker-compose. I imported [go-dockerpty](https://github.com/fgrehm/go-dockerpty) in a non academic way (and clean it up) as it was not compiling under windows and redefining stuff that are already available on `docker/docker/pkg/…`. I'll probably make a PR to the repos and vendor it if it makes it and we think it's better to do.

- [ ] Need to fix/add some integration tests.

🐸
Signed-off-by: Frederic Branczyk <fbranczyk@gmail.com>

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -161,6 +161,27 @@ func ProjectPull(p *project.Project, c *cli.Context) {
 	}
 }
 
+// ProjectRun runs a given command within a service's container.
+func ProjectRun(p *project.Project, c *cli.Context) {
+	if len(c.Args()) == 1 {
+		logrus.Fatal("No service specified")
+	}
+
+	serviceName := c.Args()[0]
+	commandParts := c.Args()[1:]
+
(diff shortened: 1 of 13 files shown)


vdemeester:
Will reimplement that with `engine-api`, should be simpler.

PR #135, opened 2016-01-03
Add flags for create command

vdemeester:
This adds the following flags (and support) for the `create` command :

- `force-recreate`
- `no-recreate`

The `--no-build` flag is shared with the `up` command and will be worked on in another PR 🐙.

- [ ] Needs integration tests

🐸

### DIFF
--- a/cli/command/command.go
+++ b/cli/command/command.go
@@ -12,6 +12,16 @@ func CreateCommand(factory app.ProjectFactory) cli.Command {
 		Name:   "create",
 		Usage:  "Create all services but do not start",
 		Action: app.WithProject(factory, app.ProjectCreate),
+		Flags: []cli.Flag{
+			cli.BoolFlag{
+				Name:  "no-recreate",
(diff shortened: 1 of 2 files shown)


cli/command/command.go
```diff
@@ -239,7 +249,7 @@ func Populate(context *project.Context, c *cli.Context) {
 
 	if c.Command.Name == "logs" {
 		context.Log = true
-	} else if c.Command.Name == "up" {
+	} else if c.Command.Name == "up" || c.Command.Name == "create" {
```
dnephin: We really need to fix this context object. It doesn't make sense to have a single context for every command.

I think we should be using something like https://github.com/spf13/cobra to handle subcommands.

vdemeester: Yeah I'm working on that, but will be another PR..

runcom: also isn't a `switch` better here?

PR #138, opened 2016-01-09
Carry #8 docker: support relative paths for volumes

vdemeester:
Carrying #8: docker: support relative paths for volumes. Rebased against master and fixed the conflicts. 🐸

Thanks @mssola for the work ! 🐼

/cc @dnephin @ibuildthecloud 

---

Consider the following yaml file:

```yaml
test:
  image: opensuse:13.2
  volumes:
    - foo:/foo
  command: sh -c 'echo Hello World'
```

docker-compose can handle the above configuration file. However this
implementation fails because it cannot handle the relative path "foo". This
commit fixes this by adding a new function `ResolvePath` to the
`ResourceLookup` interface (previously known as `ConfigLookup`) that is called
when converting the config file. The `FileConfigLookup` type implements this
function by converting relative paths to absolute paths.

Moreover, this commit also merges two for loops that were dealing with volumes.

Signed-off-by: Miquel Sabaté Solà <msabate@suse.com>

### DIFF
--- a/docker/container.go
+++ b/docker/container.go
@@ -302,7 +302,7 @@ func volumeBinds(volumes map[string]struct{}, container *dockerclient.Container)
 }
 
 func (c *Container) createContainer(imageName, oldContainer string) (*dockerclient.APIContainers, error) {
-	createOpts, err := ConvertToAPI(c.service.serviceConfig, c.name)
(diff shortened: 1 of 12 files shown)


dnephin:
LGTM

PR #148, opened 2016-02-13
Fixing pull behavior that download everything

vdemeester:
With the bump to docker v1.10.0, the pull behavior was broken, it downloads all tags for an image (which takes a loooonng time). :sweat: 

🐸


vdemeester:
hum well.. will be fixed by the switch to engine-api then..

PR #149, opened 2016-02-17
fix the testcase TestInParallelError

datawolf:
booleanMap expected to contain true values on odd number, so move
the code `booleanMap[index] = true` down to fix that.

Signed-off-by: Wang Long <long.wanglong@huawei.com>

### DIFF
--- a/utils/util_inparallel_test.go
+++ b/utils/util_inparallel_test.go
@@ -61,12 +61,13 @@ func TestInParallelError(t *testing.T) {
 	for i := 0; i < size; i++ {
 		task := func(index int) func() error {
 			return func() error {
-				booleanMap.Add(index, true)
+				booleanMap.Add(index, false)
 				t.Log("index", index)
 				if index%2 == 0 {
 					t.Log("return an error for", index)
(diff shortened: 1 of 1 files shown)


joshwget:
I think the underlying problem with the test is that `!value` is checked for odd values, but `booleanMap` doesn't contain any entries that aren't `true`. Therefore, `!value` will always evaluate to false. I think your change has the right idea, but we should add `booleanMap[index] = false` where `booleanMap[index] = true` previously was.

```
booleanMap[index] = false
if index%2 == 0 {
    return fmt.Errorf("Error with %v", index)
}
booleanMap[index] = true
```

datawolf:
@joshwget thanks, I have updated the patch.

joshwget:
Can you check this out @vdemeester? I think it's ready to be merged.

vdemeester:
@datawolf needs a rebase though :sweat_smile:

datawolf:
@vdemeester have done.

vdemeester:
It is green.
LGTM 🐸

PR #150, opened 2016-02-18
On Ctrl+c, libcompose-cli will stop the service gracefully.

### LINKED ISSUE #131: up command without detach should stop container on ctrl-c
When using `docker-compose up`, it starts all the containers in foreground (or kinda :stuck_out_tongue_closed_eyes:) ; thus when issuing `ctrl-c`, it shuts them down. `libcompose-cli` on master does not.

```bash
$ docker-compose up 
# […]
# Hitting ctrl-c
$ docker-compose ps 
    Name                   Command               State    Ports 
---------------------------------------------------------------
compose_db_1    /docker-entrypoint.sh postgres   Exit 0         
compose_web_1   nginx -g daemon off;             Exit 0 

$ libcompose-cli up
# […]
# Hitting ctrl-c
$ libcompose-cli ps
Name           Command                         State         Ports
compose_db_1   /docker-entrypoint.sh postgres  Up 5 seconds  5432/tcp
compose_web_1  nginx -g 'daemon off;'          Up 6 seconds  443/tcp, 80/tcp
```

kunalkushwaha:
BugFix #131 , On libcompose-cli up, if ctrl+c is pressed, it will
gracefully stop all containers of service.

Closes #131

Signed-off-by: Kunal Kushwaha <kushwaha_kunal_v7@lab.ntt.co.jp>

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -3,8 +3,10 @@ package app
 import (
 	"fmt"
 	"os"
+	"os/signal"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
@@ -124,7 +126,7 @@ func ProjectUp(p *project.Project, c *cli.Context) {
 	}
 
 	if !c.Bool("d") {
-		wait()
+		exitOnSignal(p, c)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


cli/app/app.go
```diff
@@ -259,3 +261,17 @@ func ProjectScale(p *project.Project, c *cli.Context) {
...
 }
+
+func exitOnSignal(p *project.Project, c *cli.Context) {
+	signalChan := make(chan os.Signal, 1)
+	cleanupDone := make(chan bool)
+	signal.Notify(signalChan, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		for _ = range signalChan {
+			fmt.Println("\nGracefully stopping...\n")
+			ProjectKill(p, c)
```
vdemeester: It should be `ProjectDown` instead of `ProjectKill` for it to be *graceful*.

vdemeester:
Looks good and simple :+1:. It would be cool if we had an integration-test (not sure how easy it would be to set it up).

kunalkushwaha:
Any thoughts for setup up for test using testing framework.

I tried to implement it, but even if `up` command is executed in go routine,
Also there is no way to pass **signal** to that go routine as process context is same as parent.

vdemeester:
@kunalkushwaha I'll take a look at the test we have in `docker/docker` :wink:

vdemeester:
LGTM 🐮
Let's merge this and iterate over for the test.

PR #165, opened 2016-03-01
Extending container events

bfosberry:
This PR makes two changes

1) Adds container ID to the container events data map
2) Polls running containers when started to notify when they stop with a ContainerStopped event

These can be easily separated if needed. 

Adding the ID is pretty low risk, since its a dynamic, and currently empty data map. We can generally ignore any error when fetching container info.

For the poller, it is higher risk, however only because of the extra goroutine being utilized. I'll still need to verify the state strings, and probably move them to be constants, or reference them directly as external constants.

Let me know how to proceed with this.

### DIFF
--- a/docker/container.go
+++ b/docker/container.go
@@ -8,6 +8,7 @@ import (
 	"math"
 	"os"
 	"strings"
+	"time"
 
 	"golang.org/x/net/context"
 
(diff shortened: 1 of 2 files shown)


bfosberry:
Will close this PR once I reproduce the work elsewhere, it depends on an event system refactor. I'll try and pull in some of these aspects and will most likely re-open a pr later for just adding ID to the container events data maps

bfosberry:
Picking this up again

PR #172, opened 2016-03-08
Check project.serviceStart return for error

joshwget:
Not checking for an error here causes libcompose to hang when circular dependencies are present. It's better to fail with a relevant error message.


ibuildthecloud:
LGTM

vdemeester:
LGTM 🐸

PR #173, opened 2016-03-10
Convert map to colon seperated string in MaporColonSlice

joshwget:
Links that are of the form `"server: foo"` seem to not be parsed correctly by candiedyaml. Instead of being parsed as a string, they are parsed as a map. This was causing errors since maps within lists aren't a valid format for links.

This converts the map back into a string and is a workaround until candiedyaml can be fixed.

### DIFF
--- a/integration/up_test.go
+++ b/integration/up_test.go
@@ -295,6 +295,37 @@ func (s *RunSuite) TestLink(c *C) {
 		fmt.Sprintf("/%s:/%s/%s", serverName, clientName, "server"),
 		fmt.Sprintf("/%s:/%s/%s", serverName, clientName, serverName),
 	}))
+
+	p = s.ProjectFromText(c, "up", `
+        server:
+          image: busybox
+          command: cat
+          stdin_open: true
+          expose:
+          - 80
+        client:
+          image: busybox
+          links:
+          - "server: foo2"
+          - server
+        `)
+
+	serverName = fmt.Sprintf("%s_%s_1", p, "server")
+
+	cn = s.GetContainerByName(c, serverName)
+	c.Assert(cn, NotNil)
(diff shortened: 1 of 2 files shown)


joshwget:
It's also worth mentioning that Docker Compose supports this syntax

vdemeester:
Cool ! One thing though, `docker-compose` keeps the space, `libcompose` with this PR does not :stuck_out_tongue_closed_eyes:.

```json
"Links": [
                "/toto_server_1:/toto_client_1/foo2",
                "/toto_server_1:/toto_client_1/server",
                "/toto_server_1:/toto_client_1/foo",
                "/toto_server_1:/toto_client_1/toto_server_1"
            ],
```

vs 

```json
"Links": [
                "/titi_server_1:/titi_client_1/ foo2",
                "/titi_server_1:/titi_client_1/foo",
                "/titi_server_1:/titi_client_1/server",
                "/titi_server_1:/titi_client_1/server_1",
                "/titi_server_1:/titi_client_1/titi_server_1"
            ],
```

The other thing missing is not in the scope of this PR, but it might be worth noting somewhere (in an issue — I'll create it tomorrow).

joshwget:
Hmm, keeping the space around might actually be a little tricky. Since candiedyaml is giving us a `map[string]interface{}`, libcompose has no way of knowing whether there were one or multiple spaces originally.

@vdemeester do you know how essential it is that the space remains? I'm not too sure what the significance of it is here.

dnephin:
Two different yaml parsers should not produce different structures for the same config. It sounds like this is a bug in candiedyaml (it's possibly a bug in pyyaml, but considering the number of users of each, I'm a lot more inclined to believe it's candiedyaml).

joshwget:
I agree @dnephin. Closing in favor of fixing in candiedyaml.

joshwget:
This should be fixed by cloudfoundry-incubator/candiedyaml#19

Issue #113, opened 2015-12-02
Quoted numbers are converted to int types in interpolation, then fail to parse 


lox:
Given:

```
sample:
  image: amazon/amazon-ecs-sample
  environment:
    test_int_as_string:  "100"
```

The result is a panic:

```
panic: interface conversion: interface is int64, not string [recovered]
	panic: interface conversion: interface is int64, not string

goroutine 1 [running]:
github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml.recovery(0xc82016e728)
	/Users/lachlan/.go/src/github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml/decode.go:83 +0xb4
github.com/docker/libcompose/project.(*MaporEqualSlice).UnmarshalYAML(0xc8201b6418, 0x600690, 0x15, 0x4123a0, 0xc8201911a0, 0x0, 0x0)
	/Users/lachlan/.go/src/github.com/docker/libcompose/project/types_yaml.go:190 +0x4cc
github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml.(*Decoder).mapping.func1(0xfa4fc8, 0xc8201b6418, 0xc82016df50, 0xc820185500)
	/Users/lachlan/.go/src/github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml/decode.go:385 +0x81
github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml.(*Decoder).mapping(0xc820185500, 0x411460, 0xc8201dc340, 0xd4)
	/Users/lachlan/.go/src/github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml/decode.go:396 +0x29f
github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml.(*Decoder).parse(0xc820185500, 0x505ee0, 0xc8201b6418, 0xd9)
	/Users/lachlan/.go/src/github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml/decode.go:206 +0x852
github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml.(*Decoder).mappingStruct(0xc820185500, 0x5894e0, 0xc8201b6300, 0xd9)
[...]

PR #174, opened 2016-03-11
Remove integer conversion when parsing configs

### LINKED ISSUE #85: Numeric attributes should also be accepted as strings

joshwget:
This was originally introduced as a workaround to fix #85, but doesn't appear to be needed anymore. I'm not exactly sure what caused it to be no longer be necessary, but the tests I wrote are still passing. Since the conversion is causing problems such as #113, it's probably best to remove it.

### DIFF
--- a/project/interpolation.go
+++ b/project/interpolation.go
@@ -3,7 +3,6 @@ package project
 import (
 	"bytes"
 	"fmt"
-	"strconv"
 	"strings"
 
 	"github.com/Sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


vdemeester:
LGTM 🐶

PR #175, opened 2016-03-11
Use project.Context interface for argument to docker.Convert

joshwget:
The function `docker.Convert` currently takes in a context of type `docker.Context`, which is an implementation of `project.Context`. It's probably best to use the latter instead so alternative implementations of `project.Context` can be used, as is the case in Rancher Compose.

### DIFF
--- a/docker/convert.go
+++ b/docker/convert.go
@@ -41,7 +41,7 @@ func isVolume(s string) bool {
 
 // ConvertToAPI converts a service configuration to a docker API container configuration.
 func ConvertToAPI(s *Service) (*ConfigWrapper, error) {
-	config, hostConfig, err := Convert(s.serviceConfig, s.context)
+	config, hostConfig, err := Convert(s.serviceConfig, s.context.Context)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


bfosberry:
Unsure how I feel about this, aren't they technically different contexts? Maybe we should have a generic "context" interface both adhere to, in a common location, or just have two interfaces?

joshwget:
I was actually incorrect about `project.Context` being an interface. `project.Context` is embedded in `docker.Context`. I still think this is a good change to make though.

`docker.Convert` does not make use of any fields from `docker.Context` outside of the embedded `project.Context`. Since it's an exposed function, it's better to be more general and not assume that the caller is using `docker.Context`. In Rancher Compose we use a different context that also embeds a `project.Context`, so we aren't able to use `docker.Convert` with `docker.Context` as an argument.

docker/convert.go
```diff
@@ -53,7 +53,7 @@ func ConvertToAPI(s *Service) (*ConfigWrapper, error) {
 	return &result, nil
 }
 
-func volumes(c *project.ServiceConfig, ctx *Context) map[string]struct{} {
+func volumes(c *project.ServiceConfig, ctx project.Context) map[string]struct{} {
```
bfosberry: Is volumes really never being used?

bfosberry:
Since this is in the docker package I'm wary of exposing to the project package more, but I have no issues with the general change, If this fixes something for you guys, I'm all for it :P

docker/convert.go
joshwget: It's being used here https://github.com/joshwget/libcompose/blob/use-project-context-interface/docker/convert.go#L138

bfosberry: Gotcha, I thought it was odd we'd have a passing build but with a function signature change

vdemeester:
As @bfosberry, I'm kinda *meh* on this change, but I don't have anything against it either…
So, I would say LGTM :stuck_out_tongue_closed_eyes: 

/cc @ibuildthecloud

ibuildthecloud:
LGTM

PR #177, opened 2016-03-13
Fixes CI failures

vdemeester:
Fixes the current CI failures 🐻.

With docker 1.10, PATH is a default environment variable set by docker. Modifying the test to only check the presence of the wanted environment variables, not all of them.

---

Use `timeout 15m` for the `test-acceptance` script 🐌.

- It's currently "freezing" on the CI, so this makes it timeout after   some time
- It's also a good way to make sure it takes not too much time..

It should bring back CI to green (and not having CI run for 60min each time too).

---

🐸

### DIFF
--- a/integration/create_test.go
+++ b/integration/create_test.go
@@ -197,7 +197,8 @@ func (s *RunSuite) TestMultipleComposeFilesOneTwo(c *C) {
 
 	c.Assert(container.Config.Image, Equals, "busybox")
 	c.Assert(container.Config.Cmd.Slice(), DeepEquals, []string{"echo", "two"})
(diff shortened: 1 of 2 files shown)


joshwget:
There's a similar environment variable comparison in `TestMultipleComposeFilesOneTwo`. It might be a good idea to switch to using `contains` there as well.

vdemeester:
@joshwget yup, I agree.

F21:
Super awesome! Thanks for getting the tests passing again! :smile:

vdemeester:
The `timeout` is really ugly but we'll iterate to find out why it times out :wink:

PR #181, opened 2016-03-26
Carry #74 Expose Interpolate for external tools

joshwget:
Changed a few new instances of `rawService` and `rawServiceMap`. As discussed in #99, interpolation will be moved into a `config` package separately from this PR.

---

It is quite useful to parse a text file in the same exact way that
compose does.  One can imagine this being useful for GUI applications or
custom parsers.

Conflicts:
	project/merge.go

### DIFF
--- a/project/merge.go
+++ b/project/merge.go
@@ -28,18 +28,21 @@ var (
 	}
 )
 
-type rawService map[string]interface{}
-type rawServiceMap map[string]rawService
+// RawService is represent a Service in map form unparsed
+type RawService map[string]interface{}
+
+// RawServiceMap is a collection of RawServices
+type RawServiceMap map[string]RawService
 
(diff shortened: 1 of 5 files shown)


project/merge.go
```diff
@@ -28,18 +28,21 @@ var (
 	}
 )
 
-type rawService map[string]interface{}
-type rawServiceMap map[string]rawService
+// RawService is represent a Service in map form unparsed
+type RawService map[string]interface{}
```
vdemeester: When moving to a `config` package, I think we should use a more *abstracted* name :stuck_out_tongue_closed_eyes:

vdemeester:
@joshwget you beat me to it :+1: :wink:

project/merge.go
joshwget: Agreed!

vdemeester:
LGTM 🐯

PR #193, opened 2016-04-08
Bump candiedyaml

joshwget:
As suggested in #173, I fixed the underlying issue in candiedyaml that was causing certain strings to be incorrectly parsed as maps. Since then I've also noticed that libcompose fails to parse strings that begin with a colon, so I put in a PR to address that issue too.

This bump includes both of the fixes, so libcompose shouldn't run into those parsing problems anymore.


vdemeester:
LGTM 🐼

PR #194, opened 2016-04-08
[WIP] new cli based on spf13/cobra

kunalkushwaha:
This patch is for review. Working code based on spf13/cobra framework.

I tried to keep the changes minimum at this point, like no changes in logrus based output. Which is required to change for making output similar to docker-compose.

Few inputs?
- Should we keep seprate file for each command ``(cli/cmd/*)`` or Single file as before. ``(command/command.go)`` .

Any other?


kunalkushwaha:
Deleting this PR for new PR

Issue #155, opened 2016-02-22
Revamp cli with spf13/cobra


vdemeester:
Switch to [spf13/cobra](https://github.com/spf13/cobra) for the CLI and make it more `docker-compose` like.
- docker-compose

``` bash
$ docker-compose up -d
Creating toto_simple_1
Creating toto_another_1
 $ docker-compose stop
Stopping toto_another_1 ... done
Stopping toto_simple_1 ... done
 $ docker-compose rm -f 
Going to remove toto_another_1, toto_simple_1
Removing toto_another_1 ... done
Removing toto_simple_1 ... done
```
- libcompose

``` bash
libcompose-cli up -d
WARN[0000] Note: This is an experimental alternate implementation of the Compose CLI (https://github.com/docker/compose) 
INFO[0000] Project [toto]: Starting project             
INFO[0000] [0/2] [simple]: Starting                     
INFO[0000] [0/2] [another]: Starting                    
INFO[0000] [1/2] [simple]: Started                      
INFO[0000] [2/2] [another]: Started 
$ libcompose-cli stop
WARN[0000] Note: This is an experimental alternate implementation of the Compose CLI (https://github.com/docker/compose) 
INFO[0000] Project [toto]: Stopping project             
INFO[0000] [0/2] [simple]: Stopping                     
INFO[0000] [0/2] [another]: Stopping                    
INFO[0000] [0/2] [another]: Stopped                     
INFO[0000] [0/2] [simple]: Stopped                      
INFO[0000] Project [toto]: Project stopped 
$ ibcompose-cli rm -f
WARN[0000] Note: This is an experimental alternate implementation of the Compose CLI (https://github.com/docker/compose) 
INFO[0000] [0/2] [simple]: Deleting                     
INFO[0000] [0/2] [another]: Deleting                    
INFO[0000] [0/2] [simple]: Deleted                      
INFO[0000] [0/2] [another]: Deleted                     
INFO[0000] Project [toto]: Deleting project             
INFO[0000] [0/2] [simple]: Deleting                     
INFO[0000] [0/2] [another]: Deleting                    
INFO[0000] [0/2] [simple]: Deleted                      
INFO[0000] [0/2] [another]: Deleted                     
INFO[0000] Project [toto]: Project deleted
```

kunalkushwaha:
Hi @vdemeester have you started working on this? If not, can I give a try?
[...]

PR #198, opened 2016-04-11
[WIP] cli based on spf13/cobra framwework

kunalkushwaha:
This patch is for review. Working code based on spf13/cobra framework.  

I tried to keep the changes minimum at this point, like no changes in logrus based output. Which is required to change for making output similar to docker-compose as per https://github.com/docker/libcompose/issues/155

Few inputs?

Should we keep separate file for each command (cli/cmd/*) or Single file as before. (command/command.go) .

Any other?

Signed-off-by: Kunal Kushwaha kushwaha_kunal_v7@lab.ntt.co.jp


script/binary
```diff
@@ -10,4 +10,4 @@ go generate
 go build \
    -ldflags="-w -X github.com/docker/libcompose/version.GITCOMMIT=`git rev-parse --short HEAD`" \
    -o bundles/libcompose-cli \
-   ./cli/main
+   ./cli/main.go
```
vdemeester: `./cli` then :wink:

cli/app/common.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+//	"github.com/docker/libcompose/newcli/app"
+	"github.com/docker/libcompose/project"
+)
+
+// Populate updates the specified project context based on command line arguments and subcommands.
+func Populate(context *project.Context, c *cobra.Command) {
```
vdemeester: We have to work to remove the need for this too, I still didn't think how (and maybe it's not in this PR's scope to keep the changes small) :stuck_out_tongue_closed_eyes:

vdemeester:
Cool @kunalkushwaha :+1: 

- I'm 100% to have each command in separate files. More files but with less lines of code and more *context* "kinda".
- As wrote in the comment, we have to get rid of the `Populate` function but not sure if it's the scope of the PR. I think we could achieve the same with the different *hooks* cobra gives us.
- We need to check and do something about the errors (like here : https://github.com/docker/libcompose/pull/198/files#diff-41d801ef80f1858d5e8e9695667e4dafR28)

You also gonna need to remove some comments and update vendoring (+ `gofmt`), but let's not worry on that for now but on the rest :wink: 

/ping @ibuildthecloud @joshwget @dnephin

kunalkushwaha:
Thanks @vdemeester 😄 

I checked using ``PersistentPreRun`` and ``PreRun`` we get rid of both Populate [[1]](https://github.com/docker/libcompose/blob/master/cli/command/command.go#L295) [[2]](https://github.com/docker/libcompose/blob/master/cli/docker/app/commands.go#L42) Functions and make code much more simpler.

Will update PR by tomorrow with changes for all commands.

kunalkushwaha:
Please review the updated code. 

Since, Factory functions were no more required, moved the ``cli/app/app.go`` function in respective cmd code.

vdemeester:
@kunalkushwaha could you split your commits to have one that does the vendoring and the others that have the actual updates on the code ? (it makes it easier to review :angel:)

cli/cmd/rm.go
```diff
@@ -0,0 +1,66 @@
...
+
+Any data which is not in a volume will be lost.`,
+	PreRun: deleteContext,
+	Run:    DeleteCmd,
+}
+
+func init() {
+	RootCmd.AddCommand(rmCmd)
+	rmCmd.SetUsageTemplate(subCommandTemplate)
+	rmCmd.Flags().BoolP("force", "", false, "Don't ask to confirm removal")
```
vdemeester: Typo here, missed a `f`:

```go
rmCmd.Flags().BoolP("force", "f", false, "Don't ask to confirm removal")
```

vdemeester:
In addition to my previous comments : this is a good star. A the end, I think we will be able to remove the context (and thus the needs for having `PreRun` and such), but it's not the scope of the PR :wink:.

kunalkushwaha:
Sorry, for late reply (My Weekend starts little earlier due to 🏯 timzone ) 

Cool, So no issues, Even I am in favour of one unit change per PR.
 
So I will do following.
1. Split the vendoring code from removal of Context.
2.  Fix issues with orginal PR of, so it should be mergeable with latest code.

cli/cmd/rm.go
```diff
@@ -21,6 +21,6 @@ Any data which is not in a volume will be lost.`,
 func init() {
 	RootCmd.AddCommand(rmCmd)
 	rmCmd.SetUsageTemplate(subCommandTemplate)
-	rmCmd.Flags().BoolP("force", "f", false, "Don't ask to confirm removal")
```
vdemeester: That worries me a bit... if it's a limitation of `cobra` we are in a *pickle*, because it will break compatibility with `docker-compose`…

PR #217, opened 2016-04-16
Add support to test against multiple daemon version

vdemeester:
You can specify against which "supported" daemon you want to be tested against (test-integration). It happens through the `DAEMON_VERSION` environment variables 🐳.

You can specify one or more value to it ; the only value supported are those in the Dockefile (`ENV DOCKER_VERSION …`). It will fail if the version is not found. Thus :

- `DAEMON_VERSION="1.9.1" make test-integration` will run integration tests against a 1.9.1 daemon
- `DAEMON_VERSION="1.10.3 1.11.0" make test-integration` will run integration tests against a 1.10.3 daemon and a 1.11.0 daemon.
- `DAEMON_VERSION="default" make test-integration` or `make test-integration` will run integration tests against the default daemon (specified in the the `Dockerfile`)

`DAEMON_VERSION="all"` or empty, it will test against all support daemon version. On the CI, `DAEMON_VERSION` is empty, thus testing against all supported version as well.

🐸

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,18 @@
 # This file describes the standard way to build libcompose, using docker
 FROM golang:1.6.0
 
-RUN apt-get update && apt-get install -y \
-    iptables \
-    build-essential \
-    --no-install-recommends
-
-# Install build dependencies
-RUN go get github.com/mitchellh/gox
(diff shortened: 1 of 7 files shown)


thaJeztah:
:8ball: :tada:

PR #222, opened 2016-04-21
Define CI jobs in Jenkinsfile

mikedougherty:
The intent of this PR is to define the current CI jobs (located here: https://jenkins.dockerproject.org/view/Libcompose/) in a programmatic and source-controlled way. Once this is landed in master we could remove those existing jobs.

Additionally because this is defined in code we can implement and re-use common code across many projects as appropriate -- For example in this changeset you can see usage of `wrappedNode`, `withChownWorkspace`, and `gitCommit` which are defined in their respective groovy files [here](https://github.com/docker/jenkins-pipeline-scripts/tree/master/vars).

In the long run it would be good to expand this to take advantage of some of the parallelism features offered by the Jenkinsfile format and e.g. run the test suites in parallel. Another future improvement would be output proper test reports that can be recorded by Jenkins.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,22 @@
+wrappedNode(label: 'linux && x86_64') {
+  deleteDir()
+  checkout scm
+  def image
+  try {
+    // TODO: split up into phases, create real test reports, etc.
+    stage "build image"
+    image = docker.build("dockerbuildbot/libcompose:${gitCommit()}")
+
+    stage "validate, test, build"
+    withEnv(["TESTVERBOSE=1", "LIBCOMPOSE_IMAGE=${image.id}"]) {
+      withChownWorkspace {
+        timeout(60) {
+          sh "make -e all"
+        }
+      }
+    }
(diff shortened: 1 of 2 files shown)


mikedougherty:
FYI the "Jenkins" build is marked as failed since the Pipeline plugin refuses to use my "untrusted" branch and instead uses a copy from the base branch (master). Since that doesn't exist, it fails. This is a security feature to prevent non-collaborators from executing arbitrary code/scripts.

vdemeester:
@mikedougherty this mean we will be able to update the 'build' definition via this file right (like adding an env' variable) ? I like that !

mikedougherty:
@vdemeester yep that's exactly right :) and that's just scratching the surface!

vdemeester:
LGTM :angel: 
I'll make a changes afterwards, but /me wants to see this *live*.

vdemeester:
@mikedougherty Hum, it started 2 jenkins build on master though :sweat_smile:

PR #223, opened 2016-04-24
Update Jenkinsfile with more steps

vdemeester:
Adding back multiple-daemon tests support too 🐮.

/cc @mikedougherty 

🐸

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # This file describes the standard way to build libcompose, using docker
-FROM golang:1.6.0
+FROM golang:1.6.2
 
 
 # virtualenv is necessary to run acceptance tests
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,22 +1,33 @@
+
 wrappedNode(label: 'linux && x86_64') {
   deleteDir()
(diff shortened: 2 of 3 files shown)


mikedougherty:
this LGTM as is, but if you want to take it a little further you can pull some of the repeated stuff out into a function. here's a quick draft/example:

```groovy
def makeTask(task, envVars=null) {
  // could send in the full list of envVars for each call or provide default env vars like this:
  withEnv((envVars ?: []) + ["LIBCOMPOSE_IMAGE=${image.id}"]) { // would need `def image` at top level of file instead of in the nested block
    withChownWorkspace {
      timeout(60) {
        sh "make -e ${task}"
      }
    }
  }
}
```

Now it can be used like so:
```
// wrappedNode { try { ...
  stage "validate"
  makeTask("validate")

  stage "test"
  makeTask("test", ["TESTVERBOSE=1", "DAEMON_VERSION=all"])

  stage "build"
  makeTask("cross-binary")
// } continue to existing `finally` block
```

vdemeester:
@mikedougherty good idea, I'll update the PR :+1:

PR #226, opened 2016-04-25
Remove goroutine for logs and use ProjectLog

vdemeester:
`Project.Up(…)` is now creating and starting the containers, no logs in goroutines. And this, `ProjectUp` (from `app.go`) is now calling `Project.Log(…)` after 🐼.

Makes it clearer (and even the logs look better).

🐸

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -132,8 +132,11 @@ func ProjectUp(p *project.Project, c *cli.Context) {
 	if err != nil {
 		logrus.Fatal(err)
 	}
-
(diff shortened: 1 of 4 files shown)


vdemeester:
/ping @mikedougherty now that the old libcompose build are disabled we are in a weird situation
- janky one are scheduled forever it seems (as the project at the other end is disabled)
- the new build for PR seems to show here only once they are done (which is not that good as it might appear green for the time beeing).

If you're on IRC, please ping me :angel:

PR #230, opened 2016-04-27
Refactoring : make argument required and less Context :)

vdemeester:
This PR is the start of a refactoring, that does the following 🐰 :

- Clean `Project.Context` from attributes that should not be there…
- … and thus making the `Project` *API* more argument driven than context driven ; it makes it easier to use it and know what is needed for each method. It also decouple a bit the different layer from the Context.
- Move `docker/builder.go` into a separate package that can work by itself and add test to it (~90% coverage)…
- … add a `test` package with a no-op implementation of the API client, used to *fake* client calls for unit tests (should help to write more unit tests).

There is still lots of room for improvements:
- I really wonder if the variadic signatures (`services ...string`) is useful ; it made me add the options and argument before the service(s) and.. I don't really like that.
- The `docker` package can be splitted in subpackage way more (for project, service & such) so that we have less and less coupling.

But this one is pretty big already so, let's say it's phase one :smile_cat: 

/cc @joshwget @dnephin @ibuildthecloud 

🐸


ibuildthecloud:
I think there's a bit too much in this one PR, but I very much like the idea of moving away from so much crap in the Context.

project/types/types.go
```diff
@@ -0,0 +1,29 @@
+package types
```
ibuildthecloud: What is the package was options.  The everything would be `options.Up`, `options.Build`.  Seems prettier to me.

ibuildthecloud:
I very much like this.  Also realized I should be signing my commits :)

project/types/types.go
vdemeester: Completely true :-)

vdemeester:
@ibuildthecloud updated with the package.

vdemeester:
@ibuildthecloud I'll take your "I very much like this" as a LGTM, and we'll iterate over it anyway :wink:

PR #231, opened 2016-04-28
Minimize usage of ContainerList

joshwget:
Container lookups based on name are currently done using `ContainerList` with a filter. However, I've actually been running into cases where this is returning a container with a different name. I think the underlying issue is that `name` isn't a valid filter (at least according to the Docker API docs). This PR moves almost all name-based container lookups to `ContainerInspect`.

An added benefit of this is that `types.ContainerJSON` is now used consistently, as opposed to the current mix of `types.Container` and `types.ContainerJSON`. Despite the name, `types.ContainerJSON` seems to be the better choice between the two.

The only remaining usage of `ContainerList` is in `container.Info`. The type of information that is provided by the function matches the response from `ContainerList` really well, so it felt right to leave it in use here.

### DIFF
--- a/docker/container.go
+++ b/docker/container.go
@@ -54,35 +54,33 @@ func NewContainer(client client.APIClient, name string, service *Service) *Conta
 	}
 }
 
-func (c *Container) findExisting() (*types.Container, error) {
-	return GetContainerByName(c.client, c.name)
-}
-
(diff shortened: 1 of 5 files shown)


vdemeester:
LGTM 🐸

PR #232, opened 2016-04-28
Fix volumes_from when the target service specifies a container name

joshwget:


vdemeester:
@joshwget oh I see what the problem was :sweat_smile: 

LGTM 🐸

PR #234, opened 2016-04-29
revise collect container name

xozrc:
containerName = /${node}/${service}${num} for swarm.
strings.SplitAfter(container.Names[0], "/") will get [/,${node},/${service}${num}], so that name[1] will get wrong container name. When the service up, the program will get crash.

### DIFF
--- a/docker/service.go
+++ b/docker/service.go
@@ -79,7 +79,7 @@ func (s *Service) collectContainers() ([]*Container, error) {
 
 	for _, container := range containers {
 		// Compose add "/" before name, so Name[1] will store actaul name.
-		name := strings.SplitAfter(container.Names[0], "/")
+		name := strings.SplitAfterN(container.Names[0], "/", 2)
 		result = append(result, NewContainer(client, name[1], s))
 	}
 


joshwget:
This is a good fix, but to me it seems a bit off to have the logic for parsing container names in libcompose. I wonder if this might be something that should be handled in engine-api. Or maybe there's a function in Docker Engine that can be used to handle this. Thoughts @vdemeester?

vdemeester:
I would definitely do the otherwise around (which is what `docker-compose` does, meaning always getting the last piece of the `/…/…`. 

```go
package main

import (
	"fmt"
	"strings"
)

func main() {
	containerNames := []string{"/ctn1", "/node1/ctn1", "/dc1/node1/ctn1"} // We could go on and on
	for _, name := range containerNames {
		// Code of the PR
		name1 := strings.SplitAfterN(name, "/", 2)[1]
		nameSplit := strings.Split(name, "/")
		name2 := nameSplit[len(nameSplit)-1]
		fmt.Println(name1, "vs", name2)
	}
}

// ctn1 vs ctn1
// node1/ctn1 vs ctn1
// dc1/node1/ctn1 vs ctn1
```
 (https://play.golang.org/p/rvcE_M3jjS)

Depending on what we want with collecting containers, I would definitly go for the 2nd one (always `ctn1`). It also seems we create the libcompose `Container` by name which worries me a bit, we should definitely use `id` internally and use only name for display :angel:.

PR #235, opened 2016-05-02
Refactoring package project, moving stuff around

vdemeester:
This PR does multiple-things as well 🐙 :

- Split `project/types.go` file into several smaller (`service.go`, `container.go` and `info.go`)
- Create a new `events` package with all the events (and related structures).
- Add a `APIProject` interface that the project implement. This interface
  is what should be used from outside the project. This makes writing
  unit-test of code using libcompose easier.
- Move some code around, mainly to not add too much noise into
  `APIProject`. `Ps`, `Port` and `Scale` method are now handled by the project
  struct (and thus can be use when using libcompose as a library), which was not the case previously.

🐸


joshwget:
LGTM

PR #237, opened 2016-05-02
Expose IsValidRemotes instead of valid remote slices

### LINKED ISSUE #95: Better way to expose ValidRemotes
Related to #73, find a better way to expose `ValidRemotes` for build*ing* ; either by exposing an better interface or by reusing something that comes from `docker/docker` pkg's / api.

Ref. https://github.com/docker/libcompose/pull/73/files#diff-9ffe418d3ead372a7fb8e2e59ea5ab15R18

vdemeester:
🐰

- Remove the public ValidRemotes struct
- Add a method IsValidRemotes that uses docker/docker/pkg/urlutil to
  check if the remote is valid or not (same way docker build does).

This should fix #95 in a clean way :angel: 

🐸

### DIFF
--- a/config/merge.go
+++ b/config/merge.go
@@ -9,19 +9,11 @@ import (
 
 	"github.com/Sirupsen/logrus"
 	yaml "github.com/cloudfoundry-incubator/candiedyaml"
+	"github.com/docker/docker/pkg/urlutil"
 	"github.com/docker/libcompose/utils"
 )
 
 var (
-	// ValidRemotes list the of valid prefixes that can be sent to Docker as a build remote location
-	// This is public for consumers of libcompose to use
-	ValidRemotes = []string{
-		"git://",
-		"git@github.com:",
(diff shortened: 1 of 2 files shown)


config/merge.go
```diff
@@ -326,3 +316,13 @@ func asString(obj interface{}) string {
 	}
 	return ""
 }
+
+// IsValidRemote checks if the specified string is a valid remote (for builds)
+func IsValidRemote(remote string) bool {
+	switch {
+	case urlutil.IsGitURL(remote), urlutil.IsURL(remote):
```
thaJeztah: why a switch? Can be an if, or even simpler;

```go
return urlutil.IsGitURL(remote) || urlutil.IsURL(remote);
```

vdemeester: @thaJeztah Lol, because… I did have more case at first :angel:. But you're absolutely right :wink:

PR #239, opened 2016-05-03
Refactor DeleteCommand to not need access to ListStoppedContainer

vdemeester:
It's now possible to pass a function that is used to request the validation (could be anything the take a list of string [container id / names] and returns true if the suppression should go on. If the function is not present, we do not ask. 🐯

I'm not sure about the name of the function though...

/cc @joshwget @ibuildthecloud @dnephin @thaJeztah 

🐸


joshwget:
LGTM! Moving `ListStoppedContainers` out of the interface will definitely be helpful :+1:

PR #241, opened 2016-05-03
Use context.Context accross the library

vdemeester:
Applications that require full plumbing of context, such as those that
rely on cancellation for resource cleanup, would be incompatible with
inconsistent plumbing 🐹.

This *connects* the net/context from the user of the libcompose API to
the engine-api and below (where request are cancellable for
example). This should also allow us to some cool and safe stuff with
concurrent goroutines and cancellation :angel: 

This will allow to handle, for example, when a user hits CTRL+C on `up` : it will cancel/close the request, etc… and could be used for more usefulness.

<del>Depends on #239 (for the refactoring on `Delete`)</del>

🐸

/cc @joshwget @ibuildthecloud @dnephin 

### DIFF
--- a/docker/container.go
+++ b/docker/container.go
@@ -65,18 +65,18 @@ func NewOneOffContainer(client client.APIClient, name string, containerNumber in
 	return c
 }
 
-func (c *Container) findExisting() (*types.ContainerJSON, error) {
-	return GetContainer(c.client, c.name)
+func (c *Container) findExisting(ctx context.Context) (*types.ContainerJSON, error) {
+	return GetContainer(ctx, c.client, c.name)
 }
 
 // Info returns info about the container, like name, command, state or ports.
-func (c *Container) Info(qFlag bool) (project.Info, error) {
-	container, err := c.findExisting()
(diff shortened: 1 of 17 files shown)


vdemeester:
@joshwget @ibuildthecloud how do you feel about this one ? :angel:

thaJeztah:
"LGTM" (not a maintainer), but interested in @joshwget and @ibuildthecloud's opinion :+1:

docker/container.go
```diff
@@ -591,7 +592,8 @@ func (c *Container) addNetNs(config *container.HostConfig, service project.Servi
 
 // ID returns the container Id.
 func (c *Container) ID() (string, error) {
-	container, err := c.findExisting()
+	// FIXME(vdemeester) container should not ask for his ID..
```
joshwget: Definitely agree with this!

joshwget:
Really makes me wish Go had default values for arguments 😄 but LGTM!

vdemeester:
yay, I'll make a followup PR to update the doc/example/README 😇

PR #243, opened 2016-05-05
Remove logrus dependency in the core of the library

vdemeester:
A library should not use a specific logger over an other (logrus here), but let the users of the library choose theirs. If it really needs to log things, it should provide a composable way to do it. 🐼

This refactoring defines a new type Logger and rename the old one into LogFormatter. A default Logger (based on fmt/log) is defined and used if nothing is used and `logrus.New()` can be used directly as logger (and that's what used for the cli)

- [ ] Not sure this is the right time for this refactoring (I feel there is more needed in some other places before)
- [ ] Maybe the default *logger* should be a NullLogger instead

### DIFF
--- a/cli/docker/app/factory.go
+++ b/cli/docker/app/factory.go
@@ -5,6 +5,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/Sirupsen/logrus"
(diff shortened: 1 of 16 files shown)


vdemeester:
I'll rework that later..

PR #254, opened 2016-05-08
Add events command

### LINKED ISSUE #215: Implement `events` command
There's a new events command for monitoring container events from the application, much like docker events. This is a good primitive for building tools on top of Compose for performing actions when particular things happen, such as containers starting and stopping.

```
Receive real time events from containers.

Usage: events [options] [SERVICE...]

Options:
    --json      Output events as a stream of json objects
```

vdemeester:
Add a `events` command and related method in `Project` 🐯.

Events command listen to real-time events from docker (filtering for the project). It's implemented using [vdemeester/docker-events](https://github.com/vdemeester/docker-events) library.

On the API side, `Events` returns a channel of Message (for engine-api) to be consumed by the client.

To do still:
- [x] Some tests (integration/unit)
- [x] Implement the `json` flag
- [x] Format date like `docker-compose` ?
- [ ] Handle error better

<del>*Note*: test should be easier to implement once `context.Context` PR is merged (#241)</del>

Closes #215.

/cc @joshwget @ibuildthecloud @dnephin 

🐸

### DIFF
--- a/integration/api_event_test.go
+++ b/integration/api_event_test.go
@@ -4,7 +4,7 @@ import (
 	"time"
 
 	"golang.org/x/net/context"
-	. "gopkg.in/check.v1"
+	check "gopkg.in/check.v1"
 
 	eventtypes "github.com/docker/engine-api/types/events"
 	"github.com/docker/libcompose/docker"
@@ -12,7 +12,8 @@ import (
 	"github.com/docker/libcompose/project/options"
 )
 
-func (s *APISuite) TestEvents(c *C) {
+func (s *APISuite) TestEvents(c *check.C) {
+	testRequires(c, not(DaemonVersionIs("1.9")))
 	composeFile := `
 simple:
   image: busybox:latest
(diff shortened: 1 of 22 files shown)


vdemeester:
ping @joshwget

integration/api_event_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	ctx, cancelFun := context.WithCancel(context.Background())
+
+	messages, err := project.Events(ctx)
+	c.Assert(err, IsNil)
+
+	go func() {
+		c.Assert(project.Up(ctx, options.Up{}), IsNil)
+		// Close after everything is done
+		time.Sleep(250 * time.Millisecond)
```
joshwget: Maybe we should switch the images to `tianon/true` here. On a slow network connection image pulling might delay the events and cause this test to fail.

project/empty.go
```diff
@@ -94,3 +95,8 @@ func (e *EmptyService) Run(ctx context.Context, commandParts []string) (int, err
 func (e *EmptyService) RemoveImage(ctx context.Context, imageType options.ImageType) error {
 	return nil
 }
+
+// Events implements Service.Events but does nothing.
+func (e *EmptyService) Events(ctx context.Context, events chan eventtypes.Message) error {
```
joshwget: I'm pretty sure I'm going to forget to add new commands here in the future. Maybe we should have some sort of API test for `EmptyService` to verify it always implements `Service` :smile:

integration/api_event_test.go
vdemeester: @joshwget yeah, on that note, I'm gonna do the same thing we do on `docker/docker` : prefetching some images before (while building the `libcompose-dev` image) so that we just pull images when we actually test the *pulling* :angel:

project/empty.go
vdemeester: @joshwget Good point, it's really easy to do : on top of the file 

```go
var _ Service = &EmptyService{}
```

integration/api_event_test.go
joshwget: Sounds like a good solution!

project/empty.go
joshwget: That's a pretty clever idea! I'll put a PR for this now before I forget.

PR #255, opened 2016-05-11
v2 parsing

joshwget:
First half of support for [version 2 of Docker Compose](https://docs.docker.com/compose/compose-file/#version-2). All v2 files should now be parsed, although no actions are taken for the new `volumes` and `networks` keys yet (the part you're likely interested in :smile: )

This is essentially a refactoring that makes projects store service configs in v2 format internally, rather than in the existing format. When parsing v1 files, they are first converted to v2 format. Processing is the same for both after this point. All existing v1 functionality should remain exactly the same.

What's left:
- Implementing volumes and networks
- Validating v2 files (although the schema is added by this PR)
- A suite of integration tests for v2 files


config/merge.go
```diff
@@ -20,56 +17,44 @@ var (
...
-
-	datas := make(RawServiceMap)
-	if err := yaml.Unmarshal(bytes, &datas); err != nil {
-		return nil, err
-	}
-
-	if err := Interpolate(environmentLookup, &datas); err != nil {
-		return nil, err
+// Merge merges a compose file into an existing set of service configs
+func Merge(existingServices *ServiceConfigs, environmentLookup EnvironmentLookup, resourceLookup ResourceLookup, file string, bytes []byte) (map[string]*ServiceConfig, map[string]*VolumeConfig, map[string]*NetworkConfig, error) {
```
vdemeester: I think we could make this a struct (the return part), something like `ProjectConfig` :angel:.
But I think that could be done in a follow-up PR :+1:

vdemeester: oh it already exists.. why don't we return it instead of these three maps ?

vdemeester:
@joshwget yay :tada: It looks good ! It doesn't seem to break anything (build is green :green_heart:), so I'm more than willing to merge it early and iterate over it (that way we can parallelize the work, for example on the `config` command, and taking care of network and such \o/).

Big LGTM :+1:

PR #256, opened 2016-05-11
Tiny tiny test refactoring

vdemeester:
- Moves `fixtures` folder to `testdata` (see http://dave.cheney.net/2016/05/10/test-fixtures-in-go)
- Move `empty.go` to `empty_test.go` as `EmptyService` is only used for test.

🐸


joshwget:
Can `EmptyService` still be used outside of libcompose if it's moved to `empty_test.go`? We actually make use of this in Rancher Compose :smile:

vdemeester:
@joshwget ah damn.. :sweat_smile: Ok I'll move it into the `test` package where there is some struct for tests (right now an `No op` API client)

vdemeester:
@joshwget updated.

vdemeester:
@joshwget hum cannot really do what I wanted to do… I would think then that rancher should define its own `EmptyService`.

vdemeester:
/ping @mikedougherty we have weird stuff happening on the build https://jenkins.dockerproject.org/job/docker/job/libcompose/branch/PR-256/2/console (and it's still not scheduled right away.. :sweat:)

mikedougherty:
@vdemeester thanks, i whitelisted the commands that were rejected (script security plugin is a little obtuse...) and ran it again.

not sure about the timing though, will try to investigate.

vdemeester:
@mikedougherty thanks :wink:

vdemeester:
@mikedougherty there is still some weirdiness :sob: (some have not been reported by jenkins it seems or fail to build :sweat:). Would it be possible to somehow add the possibility to run a job for a given PR (using gordon and `!rebuild` on irc for example ?

PR #257, opened 2016-05-12
Fix named volumes

joshwget:
Named volumes are being incorrectly resolved as relative volumes. This fixes the issue and removes the related warning, which doesn't seem to be present in Docker Compose anymore.


vdemeester:
Hum `07:58:17 FAIL: scale_test.go:10: CliSuite.TestScale` I think there is a timing issue on this test (it's not the first time it fails..)

so LGTM 🐰

PR #259, opened 2016-05-12
Add options to make parsing more customizable

joshwget:
Adds a framework to specify a set of parsing-related options. Currently it allows disabling interpolation/validation and adds hooks for preprocessing and postprocessing configs. Being able to specify a custom schema is something I'm hoping to add to this soon.

### DIFF
--- a/cli/docker/app/factory.go
+++ b/cli/docker/app/factory.go
@@ -30,5 +30,5 @@ func (p *ProjectFactory) Create(c *cli.Context) (project.APIProject, error) {
 
 	context.ProjectName = c.GlobalString("project-name")
 
-	return docker.NewProject(context)
+	return docker.NewProject(context, nil)
 }
--- a/config/merge.go
+++ b/config/merge.go
@@ -15,10 +15,18 @@ var (
 		"links",
 		"volumes_from",
 	}
+	defaultParseOptions = ParseOptions{
+		Interpolate: true,
+		Validate:    true,
+	}
 )
 
 // Merge merges a compose file into an existing set of service configs
-func Merge(existingServices *ServiceConfigs, environmentLookup EnvironmentLookup, resourceLookup ResourceLookup, file string, bytes []byte) (map[string]*ServiceConfig, map[string]*VolumeConfig, map[string]*NetworkConfig, error) {
+func Merge(existingServices *ServiceConfigs, environmentLookup EnvironmentLookup, resourceLookup ResourceLookup, file string, bytes []byte, options *ParseOptions) (map[string]*ServiceConfig, map[string]*VolumeConfig, map[string]*NetworkConfig, error) {
+	if options == nil {
+		options = &defaultParseOptions
+	}
+
 	var config Config
 	if err := yaml.Unmarshal(bytes, &config); err != nil {
 		return nil, nil, nil, err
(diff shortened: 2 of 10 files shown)


vdemeester:
Ah `golint` errors on ParseOption

config/types.go
```diff
@@ -221,3 +221,10 @@ type RawService map[string]interface{}
 
 // RawServiceMap is a collection of RawServices
 type RawServiceMap map[string]RawService
+
+type ParseOptions struct {
```
vdemeester: Missing a comment :stuck_out_tongue:.
I also wonder if it would make sence to put it in `project/options` package instead and thus would become `options.Parse`.

joshwget: I think it makes a bit more sense to me to keep it in `config`. These options should only be relevant to the parsing side, whereas everything in `project/options` seems to just apply to just project actions.

cli/docker/app/factory.go
```diff
@@ -30,5 +30,5 @@ func (p *ProjectFactory) Create(c *cli.Context) (project.APIProject, error) {
 
 	context.ProjectName = c.GlobalString("project-name")
 
-	return docker.NewProject(context)
+	return docker.NewProject(context, nil)
```
vdemeester: This is more for future (because I think it's safe to merge it that way) but I feel we are not *hugely* consistent with the way we init the `Project` : we pass a lot of *options* via `context`, and some without… The distinction is not that easy/clear… 

I think our end-goal should to have required argumemnts and a struct to pass any options that is not required (or that have some defaults). But it's not really in the PR scope.

config/merge.go
```diff
@@ -54,6 +62,14 @@ func Merge(existingServices *ServiceConfigs, environmentLookup EnvironmentLookup
 
 	adjustValues(serviceConfigs)
 
+	if options.Postprocess != nil {
+		var err error
+		serviceConfigs, err = options.Postprocess(serviceConfigs)
```
vdemeester: @joshwget do you have use cases for pre and post process ? Asking that to understand if passing it `serviceConfigs` is sufficient or passing more things (like the project or the config) in addition to the `serviceConfigs` would be better ?

vdemeester:
@joshwget one question (and a remark but not really in the scope of the PR, just want to have some opinion on it), otherwise this looks good.

docker/project.go
```diff
@@ -15,7 +15,7 @@ import (
 const ComposeVersion = "1.5.0"
 
 // NewProject creates a Project with the specified context.
-func NewProject(context *Context) (project.APIProject, error) {
+func NewProject(context *Context, parseOptions *config.ParseOptions) (*project.Project, error) {
```
marclop: Same as in #260, changing the return `project.APIProject` to `*project.Project`. I really think it's a better idea to return an abstraction of that struct since we can then provide a Mock on the tests and be able to asses correct functionality. New methods can be added to the interface if they need to be accessed from outside the package

joshwget: Good catch! Didn't mean for that to be in this PR. I'll fix and we can move discussion over to #260.

config/merge.go
joshwget: Our use case for preprocess is to convert booleans to strings in `environment` and `labels` to maintain backwards compatibility with Compose files that haven't been updated. No immediate use case for postprocess, but it seemed like something that may be handy in the near future. Adding those fields may also be helpful in the future, so I'm definitely open to it!

vdemeester: Ok, let's not add things we don't need for now and wait for user input then :angel: and let the signature like this right now :wink:

joshwget:
@marclop @vdemeester Updated to remove changing `docker.NewProject` return type

vdemeester:
LGTM 🐸
I'll run `make all` locally to make sure it builds :angel:

PR #260, opened 2016-05-12
docker.NewProject returns project struct instead of interface

joshwget:
The project struct defines some functions not available in the interface. If the interface is returned, we lose access to these when using `docker.NewProject`.

### DIFF
--- a/docker/project.go
+++ b/docker/project.go
@@ -15,7 +15,7 @@ import (
 const ComposeVersion = "1.5.0"
 
 // NewProject creates a Project with the specified context.
-func NewProject(context *Context) (project.APIProject, error) {
+func NewProject(context *Context) (*project.Project, error) {
 	if context.ResourceLookup == nil {
 		context.ResourceLookup = &lookup.FileConfigLookup{}
 	}


vdemeester:
I'm fine with that, but just out of curiosity, which methods do you need to access ? :angel:

marclop:
Why don't we just extend the interface instead of changing the return type? :hushed:

vdemeester:
@joshwget I kinda agree with @marclop : which one are needed ? it might make sense to have method for it in the interface :angel:

joshwget:
I think we use most of the public project functions. At the very least I know we use `CreateService`, `AddConfig`, and `Load`. Are these types of functions appropriate for the interface? I'm not sure where we want to draw the line for it.

vdemeester:
@joshwget I'm fine with poping them up in the interface (and pretty sure `Load` is even gonna be needed when I'll start implementing the `config` command). Let's add them in and we'll refactor if we feel the need for it :angel:

vdemeester:
@joshwget could you update the PR to add the required methods to `APIProject` instead ? :angel:

Issue #216, opened 2016-04-16
Support for depends_on options


vdemeester:
> There's a new depends_on option for specifying dependencies between services. This enforces the order of startup, and ensures that when you run docker-compose up SERVICE on a service with dependencies, those are started as well.

PR #269, opened 2016-05-20
Implement 'depends_on'

joshwget:
Implementation of the `depends_on` key that builds on how dependencies are already handled. Adding tests for this is a bit difficult without a framework for testing dependencies in general, so I figure better testing can be handled separately.

#216


vdemeester:
waw the implementation feels so simple.
Agree on the test, as much as I like them, it could be done afterwards (and might be covered by acceptance tests even)

LGTM 🐮

PR #270, opened 2016-05-20
Allow integers in map values, except in 'labels' key

joshwget:
Allows integers as values in all maps except the `labels` key, which gives an error in Docker Compose.

#266 


vdemeester:
LGTM 🐸
Green locally, merging :angel:

PR #273, opened 2016-05-22
Add a config command

vdemeester:
Add a simple config command (without any options for now) that does the same as `docker-compose config` 🐼.

It's pretty early on and the output will need tweaking. Even with tweaking it is going to be difficult to get the exact same output as docker-compose. The main point is to produce valid yaml that can be reused by both of them.

- [x] Need to add some test
- [x] Need to get rid of some empty key too (I'm not sure how, @joshwget help :sweat_smile: )
- [x] keep default value too
- [x] Need to add more flags :smile: (in this PR or not)

```
$ cat integration/assets/v2-full/docker-compose.yml 

version: "2"

volumes:
  data:
    driver: local

networks:
  front: {}

services:
  web:
    build:
      context: .
    networks:
      - front
      - default
    volumes_from:
      - other

  other:
    image: busybox:latest
    command: top
    volumes:
      - /data

$ ./bundles/libcompose-cli -p toto -f integration/assets/v2-full/docker-compose.yml config
version: "2.0"
services:
  other:
    build: {}
    command: [top]
    image: busybox:latest
    logging: {}
    volumes:
    - /data
    ulimits: {}
  web:
    build:
      context: integration/assets/v2-full
    logging: {}
    networks:
    - front
    - default
    volumes_from:
    - other
    ulimits: {}
volumes:
  data:
    driver: local
networks:
  front:
    ipam: {}
```

*Notes* : 
- it slighly differ from `docker-compose config` output, mainly on the order of the keys and the empty struct.
- some flags of the `config` command are for validating only. At the point we arrive in `project.Config` the validation is already done, so it might end up as a no-op :stuck_out_tongue_closed_eyes:.
- `command` are not printed the same way as with `docker-compose` : `command: [top]` vs `command: "top"`
- `build: .` is still possible to use with `v2` compose file, but fails miserably with libcompose.. 

🐸

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -275,6 +275,16 @@ func ProjectKill(p project.APIProject, c *cli.Context) error {
 	return nil
 }
 
+// ProjectConfig validates and print the compose file.
+func ProjectConfig(p project.APIProject, c *cli.Context) error {
+	yaml, err := p.Config()
+	if err != nil {
+		return cli.NewExitError(err.Error(), 1)
+	}
+	fmt.Println(yaml)
+	return nil
+}
+
 // ProjectPause pauses service containers.
 func ProjectPause(p project.APIProject, c *cli.Context) error {
 	err := p.Pause(context.Background(), c.Args()...)
--- a/cli/command/command.go
+++ b/cli/command/command.go
@@ -45,6 +45,15 @@ func CreateCommand(factory app.ProjectFactory) cli.Command {
 	}
 }
 
+// ConfigCommand defines the libcompose config subcommand
+func ConfigCommand(factory app.ProjectFactory) cli.Command {
+	return cli.Command{
+		Name:   "config",
+		Usage:  "Validate and view the compose file.",
+		Action: app.WithProject(factory, app.ProjectConfig),
+	}
+}
+
 // BuildCommand defines the libcompose build subcommand.
 func BuildCommand(factory app.ProjectFactory) cli.Command {
 	return cli.Command{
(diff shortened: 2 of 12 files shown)


joshwget:
As another checkbox to add, it would be nice if it didn't drop default values as well. For example, specifying `read_only: false` won't show with the command as it's implemented now.

I think the proper solution for this is for the `config` package to work exclusively with maps rather than unmarshaling to `ServiceConfig` and back. When this is done, default values are lost from the map. The map should only be converted to a `ServiceConfig` set as the last step, and optionally. This would allow the config command to preserve any default values.

joshwget:
I think moving all struct values in `ServiceConfig` to pointers to structs is the only solution for the second checkbox, but I'll look into it more.

vdemeester:
I'm gonna wait for #292 to update this one :angel:

joshwget:
@vdemeester I think I'd like to start working on some of these features. Would you mind getting this in a state to merge? (just tests passing and such, we can hold off on the other features you mention)

vdemeester:
@joshwget you mean merging as is (almost, but green) and iterate over it ?

joshwget:
Yes, exactly. I think what this PR contains now already has a lot of value for validation purposes.

vdemeester:
ok, sgtm, I'll update the evening 😇

joshwget:
Thanks! :smile:

vdemeester:
Let's wait for the CI to trigger but this should be green now :angel:

PR #280, opened 2016-05-27
Small service refactoring

vdemeester:
Small refactoring around `service.go` 🐭.

- <del>This moves `AuthLookup` in `config` package as other `*Lookup` and thus moves the field to `project.Context`</del>
- `eachContainer` now takes a list of containers and a new methods appears that does what `eachContainer` was doing (`collectContainersAndDo`). This way, we reduce some API calls for some cases (where to `collectContainers` were done in a row).

/cc @joshwget 

🐸

### DIFF
--- a/docker/service.go
+++ b/docker/service.go
@@ -63,7 +63,7 @@ func (s *Service) Create(ctx context.Context, options options.Create) error {
 	}
 
 	if len(containers) != 0 {
-		return s.eachContainer(ctx, func(c *Container) error {
+		return s.eachContainer(ctx, containers, func(c *Container) error {
 			return s.recreateIfNeeded(ctx, imageName, c, options.NoRecreate, options.ForceRecreate)
 		})
 	}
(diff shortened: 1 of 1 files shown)


joshwget:
I'm not sure I agree with moving `AuthLookup` to `config`. The other lookups are used in the config process, whereas `AuthLookup` isn't. Seems like it's more relevant to the `docker` package.

Is the only motivation for this to combine the two contexts? I think there's a pretty good case for keeping two. There's likely always going to be meta information that pertains only to the `docker` package. Moving those fields to project or config doesn't make sense for someone who wants to use libcompose but not the `docker` package.

vdemeester:
@joshwget hum, I moved it for another refactoring and it made sense to me to move it to the project `Context`. On that, I think I see what you mean but using `libcompose` without the `docker` project is gonna be harder and harder with support of `networks` and `volumes` (although it might be possible to abstract a lot) — networks are initialize at the project level (I'm creating a `docker/network` package that `project` package imports).

What would be the use case of using `libcompose` without the `docker` package ? I guess using the compose files and project methods but on another *backend* like kubernetes right ? (what you are doing with rancher ?).

joshwget:
I think we should abstract networks and volumes in the same way as we do services. We can have an interface for them in project, and then an implementation in the `docker` package. This should allow the `project` package to not depend on `docker` or any of its subprojects.

Communicating with a different backend is definitely the use case I had in mind, but there are probably some others as well. Since that backend can be pretty much anything, `project` shouldn't have any logic specific to interacting with a Docker daemon (plenty of gray areas here, of course).

vdemeester:
@joshwget ok sounds good to me, I'll orient some of my refactoring with that in mind then :angel:

vdemeester:
@joshwget Removed the AuthLookup move :wink:

joshwget:
LGTM

PR #281, opened 2016-05-27
Empty service check

joshwget:
Adds remaining functions to `EmptyService` and a check to ensure it implements `Service`. Thanks to @vdemeester for the clever method of checking.


vdemeester:
LGTM 🐮

PR #285, opened 2016-05-28
Refactoring of events command

vdemeester:
Quick refactoring of the events command/method to not tie `project` to `engine-api` types or `docker` package or `engine-api` types 🐹.

- Do not depend on engine-api event types on `project` package, and thus define a new ContainerEvent` struct in `project/events` package.
- Do not pop engine-api event types to the user of libcompose api.

/cc @joshwget :angel: 

🐸

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -1,21 +1,17 @@
 package app
 
 import (
-	"encoding/json"
 	"fmt"
 	"os"
 	"os/signal"
 	"strconv"
 	"strings"
 	"syscall"
-	"time"
 
 	"golang.org/x/net/context"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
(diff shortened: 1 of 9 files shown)


joshwget:
I don't actually think I see any issues with using `engine-api` types in `project` (or `config`). As far as I'm aware, using the data structures defined in `engine-api` doesn't force the use of a Docker backend. I think the structures are pretty necessary aspects of `config` and `project`, so if we're not using them from `engine-api` then we'll probably end up redefining almost the same things in libcompose.

vdemeester:
@joshwget true, but libcompose should not expose through its API a struct from a dependency. That is kind of a code smell : you show details of your implementation and you're really tied to your depenency. Changing the depenency should not make the user of the library to change something (and internal changes should be minimum) 😇.

That's the main reason of the refactoring.

PR #286, opened 2016-05-28
Introduce a RemoveOrphans and RuntimeProject

vdemeester:
The `RuntimeProject` interface defines project-level methods that need direct access to the *backend* runtime (i.e. docker) 🐮.

This allows to not depend on docker in `project` package and thus, making it less coupled to any runtime.

This also moves back `ClientFactory` to `docker/client` package as it is specific to the docker runtime *and* is only required in `docker` package.

/cc @joshwget @ibuildthecloud 

🐸


docker/project.go
```diff
@@ -68,3 +77,40 @@ func NewProject(context *Context, parseOptions *config.ParseOptions) (project.AP
 
 	return p, err
 }
+
+// DockerProject implements project.RuntimeProject and define docker runtime specific methods.
```
vdemeester: This might be better off in a package by itself (like `docker/project` or something) or at least a new file :angel:.

This also need unit tests :stuck_out_tongue_closed_eyes:

project/project.go
```diff
@@ -8,10 +8,7 @@ import (
 	"golang.org/x/net/context"
 
 	log "github.com/Sirupsen/logrus"
-	"github.com/docker/engine-api/types"
```
vdemeester: No more `engine-api` dependencies here :tada:

joshwget:
I didn't notice there was already code in `project` making calls to a Docker daemon :sweat: I think this is a pretty important change, so I'm going to write up some more detailed thoughts on this. I'll try to reply sometime tomorrow :smile:

vdemeester:
@joshwget ok :+1:. When implementing the `removeOrphans` I didn't really take the separation into account :sweat: my bad :sweat_smile:

joshwget:
@vdemeester No problem! :smile: 

So I think the underlying issue here is that `Project` isn't very extensible in general. Previously `Project` commands were essentially wrappers around the corresponding `Service`commands, so providing a custom `Service` implementation gave enough flexibility for most use cases. Now there's much more logic within the `Project` commands themselves. Some of the logic is a problem because it's tied closely to interacting with a Docker daemon, but other logic might need to be customized for other reasons.

Rather than making a new interface to specifically wrap calls to a Docker backend, we might be able to just reuse `APIProject`. In this case I think it would look something like this:
1. Add `RemoveOrphans` to `APIProject`
2. Provide a new implementation of `APIProject` in `docker` that extends `project.Project` and only overrides `RemoveOrphans`

This process should also allow doing something such as overriding `ListStoppedContainers`, which I've been looking to do lately.

vdemeester:
@joshwget that's a good idea too 😝

vdemeester:
@joshwget rework in that matter, introducing a `docker.Project` that embedded `project.Project` and thus implements `project.APIProject`. There is still more refactoring to do (like not having `projet.Project` in the `docker.Service` and thus be more explicit on service calls from project but it's not the scope of this PR :angel: ).

joshwget:
All looks good except moving `DefaultDependentServices` into `docker`. I don't think there's anything in it that requires a Docker backend (we use it in Rancher Compose with no issues).

joshwget:
Only the last two dependency types depend on `engine-api`, so perhaps we can split it into a `project.DefaultDependentServices` and `docker.DefaultDependentServices` which calls the former.

vdemeester:
@joshwget yep that's what I was thinking about this morning :stuck_out_tongue_closed_eyes:. I'll take a look :)

vdemeester:
@joshwget tried something (but modified the signature of `DefaultDependentServices` though, not sure I like that..

vdemeester:
Hum the *big* problem of this approach (the way I did it) is that, right now, `RemoveOrphans` from `docker.Project` won't ever be called, unless I override the `Down` method and duplicate some code… which does not scale at all…

joshwget:
Ah, you're right. It looks like embedded structs don't work how I expected them to. I think your first approach might have been the correct method. Sorry for the incorrect suggestion :sweat: 

Would it be possible to have two separate `DefaultDependentServices` functions? I think we can avoid changing the function signature if we do that rather than just moving the two helpers to `docker`.

vdemeester:
Hum but what would you put in each then ? :angel:

joshwget:
`project.DefaultDependentServices` can be the following

```go
func DefaultDependentServices(p *Project, s Service) []ServiceRelationship {
	config := s.Config()
	if config == nil {
		return []ServiceRelationship{}
	}

	result := []ServiceRelationship{}
	for _, link := range config.Links {
		result = append(result, NewServiceRelationship(link, RelTypeLink))
	}

	for _, volumesFrom := range config.VolumesFrom {
		result = append(result, NewServiceRelationship(volumesFrom, RelTypeVolumesFrom))
	}

	for _, dependsOn := range config.DependsOn {
		result = append(result, NewServiceRelationship(dependsOn, RelTypeDependsOn))
	}

	return result
}
```

and then `docker.DefaultDependentServices` can call this and append the last two dependencies. I might be missing something, but I think that would work.

vdemeester:
@joshwget ah make sence :+1: I'll update this real soon :angel:

joshwget:
Thanks @vdemeester! :smile:

vdemeester:
@joshwget Updated again, `RuntimeProject` back and `DefaultDependentServices` split :angel:

joshwget:
The name `RuntimeProject` seems a bit specific, since it might be used to override other kinds of functionality. But I think it's okay until we see that use case, so LGTM!

vdemeester:
yay :tada:

PR #287, opened 2016-05-29
Expose ImageExists in docker.Service

joshwget:
`ImageExists` is a pretty useful helper, so I think it makes sense to expose it. This also changes it to return a boolean rather than bubble up an `ImageNotFound` error.

### DIFF
--- a/docker/service.go
+++ b/docker/service.go
@@ -116,14 +116,12 @@ func (s *Service) createOne(ctx context.Context, imageName string) (*Container, 
 }
 
 func (s *Service) ensureImageExists(ctx context.Context, noBuild bool) (string, error) {
-	err := s.imageExists()
-
-	if err == nil {
-		return s.imageName(), nil
-	}
-
-	if err != nil && !client.IsErrImageNotFound(err) {
+	exists, err := s.ImageExists(ctx)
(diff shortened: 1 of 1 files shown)


docker/service.go
```diff
@@ -136,11 +134,19 @@ func (s *Service) ensureImageExists(ctx context.Context, noBuild bool) (string,
 	return s.imageName(), s.Pull(ctx)
 }
 
-func (s *Service) imageExists() error {
-	client := s.clientFactory.Create(s)
+// ImageExists returns whether or not the service image already exists
+func (s *Service) ImageExists() (bool, error) {
```
vdemeester: We should pass the `context.Context` here though.

```go
func (s *Service) ImageExists(ctx context.Context) (bool, error) {
    // […]
    _, _, err := dockerClient.ImageInspectWithRaw(ctx, s.imageName(), false)
    // […]
}
```

vdemeester:
Sounds good to me, just one comment :wink:

docker/service.go
joshwget: Good catch! :smile:

vdemeester:
LGTM 🐭

Issue #288, opened 2016-05-30
 'libcompose-cli up' creates project with undefined links


kunalkushwaha:
With reference to docker compose file as below.

``` yaml
web:
  image: ubuntu
  links:
   - redis
   - kk
redis:
  image: redis
```

``libcompose-cli up`` should fail, as ``kk`` is undefined. But it creates the project.

This behaviour is not similar to ``docker-compose`` which prints error and stop.

kunalkushwaha:
Closing the Bug, since its fix is merged.

Issue #277, opened 2016-05-25
Composefile v2: support volumes


vdemeester:
Add support for managing volumes for v2 services (meaning for projects that have `VolumeConfig` populated).

shouze:
Thank you, I'm new to the project so maybe I'll have some questions of course. I've cloned the project & digging into it ATM.

vdemeester:
@shouze no problem :wink: If you're on IRC, feel free to ping me :wink:

joshwget:
@shouze @vdemeester Have either of you started any work on this yet? I'd like to add this feature soon and I'm not sure if I'll be starting from scratch or not :smile:

vdemeester:
@shouze started to work on it :angel:

shouze:
@vdemeester ok, I can push the few code I've already produced you can cherry pick any commit if you want: https://github.com/shouze/libcompose/compare/master...277-support-volumes.

joshwget:
@shouze Is what you have in a complete enough state that it could be put in master? If that's the case maybe we can just get that in and iterate from there.

If not, would you mind if I cherry-pick your commit and finish off the remaining work?

shouze:
@joshwget not complete enough, would prefer a cherry-pick. Also you probably may have to make 2 PR:
- The first one with the last commit I've pushed to expose yaml volume config like @vdemeester have done with network's one in https://github.com/docker/libcompose/pull/292.
- The second one with volume support implemented.

PR #292, opened 2016-06-01
Fixes some v2 yaml parsing 🐹

### LINKED ISSUE #291: build: path not supported
```yaml
version: '2'
services:
  app:
    build: app
    image: app
```

won't parse correctly due to [this line](https://github.com/docker/libcompose/blob/master/config/merge_v2.go#L189) even though it is correct v2 [syntax](https://docs.docker.com/compose/compose-file/#build)

vdemeester:
This PR does two things:

#### Store config version in project 🐮 

This is needed for `networks` and `volumes` support as we are gonna need to know the version of the config file to act based on it.

#### Some `yaml` refactoring/magic and `config` updates 🐹

Doing several things: 
- Split into smaller files (with test) to make it easier to read.
- Add a `Build` type that handle some Build magic
- Add an `External` type that handle some external magic (networks, volumes)
- Add `Networks`/`Network` type that handle some Networks magic
- Refactor the `Ulimit` types to be simpler and not user reflection

Also adds a `merge_fixtures_test.go` and `testdata` folder with some existing compose file to make sure it parses it well (should serve as quick regression tests in the future too).

/cc @joshwget (master of the yaml :joy: :stuck_out_tongue_closed_eyes:)

Fixes  #291
Needed for #277 and #288 implementations

🐸 🐸


yaml/build.go
```diff
@@ -0,0 +1,101 @@
...
+				continue
+			}
+		}
+	default:
+		return fmt.Errorf("Failed to unmarshal Build: %#v", value)
+	}
+	return nil
+}
+
+func handleBuildArgs(value interface{}) (map[string]string, error) {
```
joshwget: Maybe we shouldn't make these two functions specific to build arguments. We'll probably have to do something similar for logging arguments, so maybe we can move these to a `utils.go`?

yaml/network.go
```diff
@@ -0,0 +1,97 @@
...
+				network.Aliases = []string{}
+				for _, alias := range aliases {
+					network.Aliases = append(network.Aliases, alias.(string))
+				}
+			case "ipv4_address":
+				network.IPv4Address = mapValue.(string)
+			case "ipv6_address":
+				network.IPv6Address = mapValue.(string)
+			default:
+				// Ignorer unknown keys ?
```
joshwget: Validation (once it's done for v2) should complain about unknown keys :smile:

joshwget:
Returning the config version is definitely a good idea. Awesome work with all of the new types. Hopefully that wasn't too tedious!

Just one comment, but LGTM other than that.

yaml/build.go
vdemeester: Hum right now `logging` key works fine, but I'll keep that in mind because yeah, that's a pretty generic way of handling a map :angel:

Issue #296, opened 2016-06-05
Environment variables specified as dictionary keys w/o value get set to <nil>


janeczku:
docker-compose.yml:

```YAML
testing:
  image: ubuntu:14.04
  environment:
    FOO:
```

> $ FOO=BAR rancher-compose up

Result:

```YAML
testing:
  environment:
    FOO: <nil>
  image: ubuntu:14.04
```

> $ FOO=BAR libcompose-cli up

Result:

```YAML
ERRO[0000] Failed to unmarshall: Cannot unmarshal '<nil>' of type <nil> into a string value
testing:
  environment:
    FOO: null
  image: ubuntu:14.04
```

Expected:

Environment variables with only a key are resolved to their values on the machine Compose is running on (https://docs.docker.com/compose/compose-file/#environment)

joshwget:
This was fixed by #297 and should be working as expected on master. Ping @vdemeester.

PR #297, opened 2016-06-05
Fixes parsing of environment dictionary keys

janeczku:
https://github.com/docker/libcompose/issues/296

Verified that this fixes parsing environment dictionary.


joshwget:
I think keys with missing values should be converted to just the key name (`VAR`, not `VAR=`). This matches how the feature works when environment variables are specified as an array. Thanks for catching this issue! :smile:

janeczku:
@joshwget Updated the PR to use just the key name without separator (`=`).

joshwget:
Awesome! LGTM.

vdemeester:
I'll check the build locally (we're having some trouble with the CI… it takes time to run and it's failing for some weird reason :sweat: )

vdemeester:
LGTM :wink:

PR #302, opened 2016-06-09
Remove --lines flag for logs command

joshwget:
This flag isn't actually implemented, so it's probably confusing to users when it's shown in help. The name is wrong as well, since Docker Compose uses `--tail` for this.


vdemeester:
@joshwget oh boy, I did add it without implementing it… my bad :sweat_smile:

joshwget:
No problem! I think I'll implement `tail` pretty soon :smile:

PR #304, opened 2016-06-09
Remove project event logging

joshwget:
Switch to only showing service events, which makes the output closer to Docker Compose. There are also scenarios when showing project logs results in some confusing behavior. For example, calling start on a subset of all services for a project will result in a log "Project started", which doesn't seem quite right in this case.


vdemeester:
LGTM 🐯

Issue #312, opened 2016-06-22
build args don't get populated


dansteen:
Hey!

Thanks so much for your work on this!  I noticed that given the following config:
```
version: '2'
services:
  org-api.app.local:
    build:
      context: .
      args:
        - VAULT_TOKEN
        - APP_VERSION
```

The build args don't get populated into the Service config.  This works in docker-compose, and it reads the values of those args from ENV variables per  https://docs.docker.com/compose/compose-file/#args:
```
Build arguments with only a key are resolved to their environment value on the machine Compose is running on.

```

I did notice that if I changed it to this:
```
version: '2'
services:
  org-api.app.local:
    build:
      context: .
      args:
        VAULT_TOKEN: bob
        APP_VERSION: dave
```

Then it would pass those values into the Service config, but it would not set the corresponding args in the actual Docker build.  I'm not sure if it makes a difference, but in the Dockerfile, all the args have default values set.

I did find this: https://github.com/docker/libcompose/pull/307 which seems like it would be relevant, but for environments rather than args.

Thanks!

joshwget:
Thanks! I didn't realize this was an option for build arguments.

If you're at all interested in taking a stab at this yourself, this should be a pretty easy fix by mimicking the [similar functionality in environment variable lookups](https://github.com/docker/libcompose/blob/master/project/project.go#L116).

dansteen:
Ok.  I'll give it a go.  Thanks!

PR #313, opened 2016-06-23
Added in compatibility with dockers handling of build args

dansteen:
Adjusts the handling of build args per https://github.com/docker/libcompose/issues/312

### DIFF
--- a/yaml/build.go
+++ b/yaml/build.go
@@ -3,6 +3,7 @@ package yaml
 import (
 	"fmt"
 	"strconv"
+	"strings"
 )
 
 // Build represents a build element in compose file.
@@ -63,20 +64,27 @@ func handleBuildArgs(value interface{}) (map[string]string, error) {
 	case map[interface{}]interface{}:
 		return handleBuildArgMap(v)
 	case []interface{}:
-		args = map[string]string{}
-		for _, elt := range v {
-			uniqArgs, err := handleBuildArgs(elt)
-			if err != nil {
-				return args, nil
-			}
-			for mapKey, mapValue := range uniqArgs {
-				args[mapKey] = mapValue
-			}
-		}
-		return args, nil
+		return handleBuildArgSlice(v)
 	default:
 		return args, fmt.Errorf("Failed to unmarshal Build args: %#v", value)
 	}
+}
+
+func handleBuildArgSlice(s []interface{}) (map[string]string, error) {
+	var args = map[string]string{}
+	for _, arg := range s {
+		// check if a value is provided
+		switch v := strings.SplitN(arg.(string), "=", 2); len(v) {
+		case 1:
+			// if we have not specified a a value for this build arg, we assign it an ascii null value and query the environment
+			// later when we build the service
+			args[v[0]] = "\x00"
+		case 2:
+			// if we do have a value provided, we use it
+			args[v[0]] = v[1]
+		}
+	}
+	return args, nil
 }
 
 func handleBuildArgMap(m map[interface{}]interface{}) (map[string]string, error) {
--- a/project/project.go
+++ b/project/project.go
@@ -119,6 +119,23 @@ func (p *Project) CreateService(name string) (Service, error) {
 		}
 
 		config.Environment = parsedEnv
+
+		// check the environment for extra build Args that are set but not given a value in the compose file
+		for arg, value := range config.Build.Args {
+			if value == "\x00" {
+				envValue := p.context.EnvironmentLookup.Lookup(arg, name, &config)
+				// depending on what we get back we do different things
+				switch l := len(envValue); l {
+				case 0:
+					delete(config.Build.Args, arg)
+				case 1:
+					parts := strings.SplitN(envValue[0], "=", 2)
+					config.Build.Args[parts[0]] = parts[1]
+				default:
+					return nil, fmt.Errorf("Tried to set Build Arg %#v to multi-value %#v.", arg, envValue)
+				}
+			}
+		}
 	}
 
 	return p.context.ServiceFactory.Create(p, name, &config)
--- a/yaml/build_test.go
+++ b/yaml/build_test.go
@@ -100,8 +100,21 @@ args:
 		{
 			yaml: `context: .
 args:
-  - buildno: 1
-  - user: vincent`,
+  - buildno
+  - user`,
+			expected: &Build{
+				Context: ".",
+				Args: map[string]string{
+					"buildno": "\x00",
+					"user":    "\x00",
+				},
+			},
+		},
+		{
+			yaml: `context: .
+args:
+  - buildno=1
+  - user=vincent`,
 			expected: &Build{
 				Context: ".",
 				Args: map[string]string{


yaml/build.go
```diff
@@ -63,22 +65,31 @@ func handleBuildArgs(value interface{}) (map[string]string, error) {
...
 	}
 }
 
+func handleBuildArgSlice(s []interface{}) (map[string]string, error) {
+	var args = map[string]string{}
+	for _, arg := range s {
+		// check if a value is provided
+		switch v := strings.SplitN(arg.(string), "=", 2); len(v) {
+		case 1:
+			// we don't use EnvironmentLookup here as docker does not use values in the env_file or environment config stanza for this
```
vdemeester: "as docker-**compose** does not …"

vdemeester:
@dansteen you need to sign-off your commit (`git commit --sign-off --amend …`)

vdemeester:
@dansteen could you squash the commits too :angel: :stuck_out_tongue_closed_eyes:

yaml/build.go
```diff
@@ -63,22 +65,31 @@ func handleBuildArgs(value interface{}) (map[string]string, error) {
...
+			envValue := os.Getenv(v[0])
```
joshwget: The problem with using `os.Getenv` is that it can't be overridden to perform lookups in a customized way. We have an [existing structure](https://github.com/docker/libcompose/blob/master/config/types.go#L10) that solves this problem.

`EnvironmentLookup` is only available in the `Project` struct, so I think some of the processing code will have to be moved to the `project` package. Specifically, we do processing for the `environment` key in `CreateService`, so it probably makes sense for this to be in `CreateService` as well.

dansteen: So, docker-compose itself doesn't read build "Args" values from the environments stanza or the env_file entry.   Is that something that you would want to support for libcompose?   Env values and build Args are handled very differently in both docker itself and docker-compose.   Having libcompose populate build Args from Env values that are specified in environments or env_file would be a significant deviation from the standard behavior...

dansteen: And then it occurred to me that you may have just meant generically that you want to be able to populate these values more dynamically.   I think you are right.

vdemeester:
@dansteen build errors

```
08:55:06 + go test -race -cover -coverprofile=cover.out github.com/docker/libcompose/yaml
08:55:10 --- FAIL: TestUnmarshalBuild (0.00s)
08:55:10 panic: interface conversion: interface is map[interface {}]interface {}, not string [recovered]
08:55:10 	panic: interface conversion: interface is map[interface {}]interface {}, not string [recovered]
08:55:10 	panic: interface conversion: interface is map[interface {}]interface {}, not string
08:55:10 
08:55:10 goroutine 6 [running]:
08:55:10 panic(0x9409e0, 0xc82000b140)
08:55:10 	/usr/local/go/src/runtime/panic.go:481 +0x3ff
08:55:10 testing.tRunner.func1(0xc8200142d0)
08:55:10 	/usr/local/go/src/testing/testing.go:467 +0x22a
08:55:10 panic(0x9409e0, 0xc82000b140)
08:55:10 	/usr/local/go/src/runtime/panic.go:443 +0x521
08:55:10 github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml.recovery(0xc82008dcf8)
08:55:10 	/go/src/github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml/decode.go:83 +0xce
08:55:10 panic(0x9409e0, 0xc82000b140)
08:55:10 	/usr/local/go/src/runtime/panic.go:443 +0x521
08:55:10 github.com/docker/libcompose/yaml.handleBuildArgSlice(0xc8200d5600, 0x2, 0x2, 0xc82008d618, 0x0, 0x0)
08:55:10 	github.com/docker/libcompose/yaml/_test/_obj_test/build.go:105 +0x183
08:55:10 github.com/docker/libcompose/yaml.handleBuildArgs(0x88c180, 0xc8200d5620, 0x2, 0x0, 0x0)
08:55:10 	github.com/docker/libcompose/yaml/_test/_obj_test/build.go:92 +0x161
08:55:10 github.com/docker/libcompose/yaml.(*Build).UnmarshalYAML(0xc82010e2a0, 0xa0d430, 0x15, 0x89d100, 0xc82010e360, 0x0, 0x0)
08:55:10 	github.com/docker/libcompose/yaml/_test/_obj_test/build.go:62 +0x630
08:55:10 github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml.(*Decoder).mapping.func1(0x7f3aec246b08, 0xc82010e2a0, 0xc82008d948, 0xc820106e00)
08:55:10 	/go/src/github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml/decode.go:385 +0xa6
08:55:10 github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml.(*Decoder).mapping(0xc820106e00, 0x89b240, 0xc8200dd110, 0x194)
08:55:10 	/go/src/github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml/decode.go:396 +0x2a1
08:55:10 github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml.(*Decoder).parse(0xc820106e00, 0x93e1a0, 0xc82010e2a0, 0x16)
08:55:10 	/go/src/github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml/decode.go:206 +0xc20
08:55:10 github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml.(*Decoder).document(0xc820106e00, 0x93e1a0, 0xc82010e2a0, 0x16)
08:55:10 	/go/src/github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml/decode.go:182 +0x1e2
08:55:10 github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml.(*Decoder).Decode(0xc820106e00, 0x93e1a0, 0xc82010e2a0, 0x0, 0x0)
08:55:10 	/go/src/github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml/decode.go:133 +0x2d4
08:55:10 github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml.Unmarshal(0xc82000b0c0, 0x31, 0x40, 0x93e1a0, 0xc82010e2a0, 0x0, 0x0)
08:55:10 	/go/src/github.com/docker/libcompose/vendor/github.com/cloudfoundry-incubator/candiedyaml/decode.go:102 +0x134
08:55:10 github.com/docker/libcompose/yaml.TestUnmarshalBuild(0xc8200142d0)
08:55:10 	/go/src/github.com/docker/libcompose/yaml/build_test.go:116 +0x714
08:55:10 testing.tRunner(0xc8200142d0, 0xc15578)
08:55:10 	/usr/local/go/src/testing/testing.go:473 +0xdd
08:55:10 created by testing.RunTests
08:55:10 	/usr/local/go/src/testing/testing.go:582 +0xae3
08:55:10 FAIL	github.com/docker/libcompose/yaml	0.017s
08:55:10 
08:55:10 Tests failed: github.com/docker/libcompose
```

yaml/build.go
dansteen: Ok.  I've updated things so that the main processing is still done in yaml/build.go, but it sets empty strings as map values (indicated by an ASCII null)  if the build Arg *value* is not provided in the compose file.   Then, when the service is created it will use ```EnvironmentLookup``` to see if there is a value for that Arg available in the environment.  I think it works nicely to delay looking for the value in the environment until the service is created as you may want to take some action between reading in the compose file and starting the services.

dansteen:
In the test failure above, the following lines are failing in yaml/build_test.go:
```
    {
      yaml: `context: .
args:
  - buildno: 1
  - user: vincent`,
      expected: &Build{
        Context: ".",
        Args: map[string]string{
          "buildno": "1",
          "user":    "vincent",
        },
      },
    },
  }
```

Those lines test a config with the syntax
```
build:
      context: .
      args:
        - buildno: 1
        - user: vincent
```
but this is not a valid docker-compose file syntax and docker-compose itself fails on this.  You can either do this:
```
build:
      context: .
      args:
        buildno: 1
        user: vincent
```

or this:
```
build:
      context: .
      args:
        - buildno=1
        - user=vincent
```

or the syntax we are adding, but not the one this test is looking for.  With your permission, I have removed that test, and added in tests for the new syntax options.

vdemeester:
@dansteen ah you're right about the test :sweat_smile: 
Overall LGTM 👼
/cc @joshwget

joshwget:
The method of using `\x00` as a temporary is clever, but I think it would be better to refactor build arguments to be a `[]string`. It's done that way for environment variables and the logic seems a bit more straightforward there (plus we might be able to reuse some existing code).

dansteen:
do you mean you want Build.Args to be a []string of "ARG=VALUE" format (similar to environment variables) or do you mean that you want Build.Args to be a map[string][]string, and then []string can either contain a single value or be empty?

joshwget:
The former (`Build.Args` should be a `[]string`)

dansteen:
So, there are three possible value types here:

ARG="VALUE" - set's the arg to "VALUE"
ARG="" - sets the arg to blank (this effectively clears the arg from the default that is in the dockerfile)
ARG - pulls the value of arg from the environment

Using a []string how would you want to represent this?

We could do something like the following:

"ARG=VALUE" 
"ARG=" 
"ARG"

Where the presence of the equals sign at the end of the value indicates that it has been set and is empty, whereas the lack of an equals sign at the end means that we pull the value from the environment.  But this is really just another form of using a "magic" value.   And I think it's actually worse since it's much more subtle, whereas \x00 is pretty explicit.   We also lose the expressiveness that a map gives us.  

One alternative is to use a struct:
```
{
  Value string
  SetFromEnv bool
}
 ```
but that seems to add a lot of overhead for this...

joshwget:
Actually, you're right. I figured engine-api would accept build arguments as a `[]string` similar to how they accept environment. It looks like it's actually passed as a `map[string]string`, so your method would save us from having to perform that conversion. LGTM!

PR #316, opened 2016-06-28
Refactoring the Container struct…

vdemeester:
… and thus Service too (with some other stuff) 🐍.

Overall the struct is way lighter, less intelligent *and* make way less
call. A `Container` now holds only a client, name, id and an actual
`types.ContainerJSON` docker container struct.

- This removes service and `Project`/`events.Notifier`.
- There is several way to create a `Container`:
  - Create an actual container with `CreateContainer`
  - Create a container from an existing one, with either `New` or
   `NewInspected`.
- Thus, removing any `Create*` method (use `CreateContainer` instead).
- Move *service* related method into services (`Recreate`, `Delete`,
  `OutOfSync`, anything related to `Create`).
- Remove `findExisting` and calls, the same with `withContainer` which
  are not needed anymore.
- Create a `service_create.go` file with service container creation
  code (because it was a lot)
- Update `Service` methods to use this updated `Container` struct.
- Makes `Scale` on service less flaky but not doing it in parallel (for
  now at least)

There is some improvements to come next :
- `Container` could (and should) be on its own `docker/container`
  package.
- `Service` will have to be refactor next (by removing dependencies on
  project).
- Move `Service` into its on `docker/service` package and split
  `service.go` into several smaller files.

*Note that `Service` and `Container` need an overall `godoc` rework, it's on my TODO :angel:*

/cc @joshwget 

🐸

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -303,7 +303,7 @@ func ProjectScale(p project.APIProject, c *cli.Context) error {
 
 	err := p.Scale(context.Background(), c.Int("timeout"), servicesScale)
 	if err != nil {
-		return cli.NewExitError(err.Error(), 1)
+		return cli.NewExitError(err.Error(), 0)
 	}
 	return nil
 }
--- a/docker/container.go
+++ b/docker/container.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"math"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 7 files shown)


cli/app/app.go
```diff
@@ -299,7 +299,7 @@ func ProjectScale(p project.APIProject, c *cli.Context) error {
 
 	err := p.Scale(context.Background(), c.Int("timeout"), servicesScale)
 	if err != nil {
-		return cli.NewExitError(err.Error(), 1)
+		return cli.NewExitError(err.Error(), 0)
```
vdemeester: This is needed because now I actually return an error in case of *scaling with a containerName set*… This is an ugly fix to make it work like `docker-compose` but I really think `compose` should return an exit code of `1` in any error case.

docker/container.go
```diff
@@ -16,69 +17,59 @@ import (
...
-	projectName     string
-	containerNumber int
-	oneOff          bool
-	eventNotifier   events.Notifier
-	loggerFactory   logger.Factory
-	client          client.APIClient
-
-	// FIXME(vdemeester) Remove this dependency
-	service *Service
+	// FIXME(vdemeester) Replace with ContainerClient with engine-api vendor update
```
vdemeester: The latest `engine-api` split `client.APIClient` into several interface (and use composition for `client.APIClient`). Using `client.ContainerClient` here will :
- make sure we just do container stuff here.
- make it easier to unit test (as we need a smaller *fake* client to implement)

docker/container.go
```diff
@@ -16,69 +17,59 @@ import (
...
-	c := NewContainer(client, name, containerNumber, service)
-	c.oneOff = true
-	return c
+// New creates a container struct with the specified client, id and name
+func New(ctx context.Context, client client.APIClient, id, name string) (*Container, error) {
+	container, err := GetContainer(ctx, client, id)
+	if err != nil {
+		return nil, err
+	}
+	return &Container{
```
vdemeester: Probably should call `NewInspected` here.

docker/container.go
```diff
@@ -126,144 +117,64 @@ func name(names []string) string {
...
 	return nil
 }
 
+// Kill kill the container.
+func (c *Container) Kill(ctx context.Context, signal string) error {
+	return c.client.ContainerKill(ctx, c.container.ID, signal)
+}
+
 // IsRunning returns the running state of the container.
+// FIXME(vdemeester): remove the nil error here
```
vdemeester: keeping the signature for now to not have too much noise on the diff, but will definitely do after.

docker/container.go
```diff
@@ -380,225 +286,19 @@ func holdHijackedConnection(tty bool, inputStream io.ReadCloser, outputStream, e
...
-	if err != nil {
-		return nil, err
-	}
-
-	config.NetworkMode = container.NetworkMode("container:" + id)
-	return config, nil
-}
-
 // ID returns the container Id.
 func (c *Container) ID() (string, error) {
```
vdemeester: Same here (I missed a `FIXME` comment :sweat:)

docker/service.go
```diff
@@ -72,21 +71,45 @@ func (s *Service) Create(ctx context.Context, options options.Create) error {
...
+	var err error
+
+	if s.serviceConfig.ContainerName != "" {
+		if count > 1 {
+			logrus.Warnf(`The "%s" service is using the custom container name "%s". Docker requires each container to have a unique name. Remove the custom name to scale the service.`, s.name, s.serviceConfig.ContainerName)
+		}
+		namer = NewSingleNamer(s.serviceConfig.ContainerName)
+	} else {
+		client := s.clientFactory.Create(s)
+		namer, err = NewNamer(ctx, client, s.project.Name, s.name, false)
```
vdemeester: I'm still debating on wether `Namer` should talk to the API (which it does right now) or not…

docker/container.go
```diff
@@ -16,71 +17,60 @@ import (
...
-	loggerFactory   logger.Factory
-	client          client.APIClient
-
-	// FIXME(vdemeester) Remove this dependency
-	service *Service
+	// FIXME(vdemeester) Replace with ContainerClient with engine-api vendor update
+	client    client.APIClient
+	name      string
+	id        string
+	container *types.ContainerJSON
```
joshwget: This name might be confusing to people who are unfamiliar with the distinction between our `Container` struct and that of engine-api. Just a thought, but maybe we can switch to a more explicit name here? I'll admit, nothing better is coming to mind at the moment...

vdemeester: Yep, I was think `innerContainer` or `apiContainer` or `dockerContainer` but all those felt a little weird :sweat:

vdemeester: Maybe `containerJSON` in fact :angel:

joshwget: `apiContainer` is my favorite of the bunch, but I don't think the others are bad either!

joshwget:
This is a pretty nice improvement in readability and I really like that logic is being shifted from `Container` to `Service`.

So far everything looks good to me, but it's also difficult to predict the effects this will have when using libcompose as a library. It would be great if we could start adding API tests along with refactoring, especially since it sounds like there's plenty of refactoring to do.

vdemeester:
> So far everything looks good to me, but it's also difficult to predict the effects this will have when using libcompose as a library. It would be great if we could start adding API tests along with refactoring, especially since it sounds like there's plenty of refactoring to do.

It depends on how `libcompose` was used :angel:. If it's only used *on the surface* (like I do on `libkermit`), this doesn't change a thingk But it defenitely could. These refactoring aim to make it better and simpler for libcompose to be used as a library and even smallest part (like why not only use the `Container` struct, etc…).

Anyway I completely agree we should write more API tests, and more tests in general. My ideas are the following :

- Any package should be small and heavily tested. If it's not easy to write test, then it needs refactoring :angel:.
- There should be way more unit tests than integration tests. And Integration tests on `libcompose` should focus on the API side. (*Note: I'm thinking of trying out `convey` as a lib to write tests, as an experiment :angel:*)
- We should have almost no integration tests for the `cli` side as they should be handled by the acceptance tests suite comming from `docker/compose` :wink:

joshwget:
That all sounds great! Relying on the Docker Compose acceptance tests for the CLI is a really good idea.

Looks like you need a rebase here :smile:

vdemeester:
@joshwget ah I was sure I rebased.. I just forgot to push :sweat_smile:

PR #320, opened 2016-07-07
Move stopped container handling to cli/app package

joshwget:
Some logic around the `rm` CLI command is currently in `project.Delete`. This makes `project.Delete` difficult for general library usage. Move this logic to the `cli/app` package.

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -234,18 +234,22 @@ func ProjectDelete(p project.APIProject, c *cli.Context) error {
 		RemoveVolume: c.Bool("v"),
 	}
 	if !c.Bool("force") {
-		options.BeforeDeleteCallback = func(stoppedContainers []string) bool {
-			fmt.Printf("Going to remove %v\nAre you sure? [yN]\n", strings.Join(stoppedContainers, ", "))
-			var answer string
-			_, err := fmt.Scanln(&answer)
-			if err != nil {
-				logrus.Error(err)
-				return false
-			}
-			if answer != "y" && answer != "Y" {
-				return false
-			}
-			return true
+		stoppedContainers, err := p.ListStoppedContainers(context.Background(), c.Args()...)
+		if err != nil {
+			return cli.NewExitError(err.Error(), 1)
+		}
+		if len(stoppedContainers) == 0 {
+			fmt.Println("No stopped containers")
+			return nil
+		}
+		fmt.Printf("Going to remove %v\nAre you sure? [yN]\n", strings.Join(stoppedContainers, ", "))
+		var answer string
+		_, err = fmt.Scanln(&answer)
+		if err != nil {
+			return cli.NewExitError(err.Error(), 1)
+		}
+		if answer != "y" && answer != "Y" {
+			return nil
 		}
 	}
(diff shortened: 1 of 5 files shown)


cli/app/app.go
```diff
@@ -234,18 +234,22 @@ func ProjectDelete(p project.APIProject, c *cli.Context) error {
 		RemoveVolume: c.Bool("v"),
 	}
 	if !c.Bool("force") {
-		options.BeforeDeleteCallback = func(stoppedContainers []string) bool {
```
vdemeester: Hum, I would have kept the `BeforeDeleteCallback` (which might have the wrong name now I think of it…) that could allow interactive stuff from the lib… but that's not really a big deal and that might complicate thing where there is no need to, so ok :angel:

project/interface.go
```diff
@@ -37,6 +37,7 @@ type APIProject interface {
 	CreateService(name string) (Service, error)
 	AddConfig(name string, config *config.ServiceConfig) error
 	Load(bytes []byte) error
+	ListStoppedContainers(ctx context.Context, services ...string) ([]string, error)
```
vdemeester: I would prefer to have a `ListContainer(s)` with some kind of filter or options. This would make it more generic and add ways to do it (I'm even needing this in `libkermit/compose` for example :angel: ).

joshwget:
Adding more container logic to `project

project/interface.go
joshwget: Adding more container logic to `project` makes it a bit trickier to use without the `docker` package. Certain container operations are difficult to implement with orchestration systems.

For example, I've never really been able to do a proper implementation of `IsRunning`. As a result, `Delete` didn't really work since it'll always say that there are no containers stopped.

I'm definitely not opposed to adding functions to the `Container` interface, but it would be helpful to gracefully handle the case where they're not implemented :smile:

cli/app/app.go
joshwget: I think if `ListStoppedContainers` is exposed then the same functionality is still available. A library user can call `ListStoppedContainers` and then just perform and "before" operations before making the call to `Delete`.

vdemeester: @joshwget yep, realized this while writing the comment :sweat_smile:

project/interface.go
vdemeester: My point is more, if we had something about `Container` in the `Project` interface, I would prefer something a little bit more *abstract* or *useful* than just having the `StoppedContainer`. Maybe the `Container` term is too technical though but it's only returning a list of ids — it doesn't feel too tied to the `docker` package to me :stuck_out_tongue_closed_eyes:.

```go
type ContainerFilter struct {
    state string
}

func (p *Project) Containers(ctx context.Context, filters ContainerFilters) {
   // […]
}
```

joshwget: Are you okay with merging this as-is and then iterating on container listing functionality later? It seems separate from this, so maybe we can address it in a different PR.

vdemeester: Why not, but if we merge as-is, the follow-up have to be in 0.3.0 (as it will update the signature 👼). I can take care of it :stuck_out_tongue_closed_eyes:

vdemeester:
LGTM 🐹

PR #321, opened 2016-07-07
Check project networks non-nil before initializing

joshwget:


vdemeester:
@joshwget thanks, I saw that :scream: 
LGTM 🐸

PR #322, opened 2016-07-08
Expose builder.createTar

joshwget:
Exposes the `createTar` function in `docker/builder/builder.go`. We make use of this function and it was exposed prior to #230.

On a related note, there is some code in `docker/builder` that can probably be moved to a separate package. Some of the code doesn't depend on a Docker daemon (such as `createTar`) and it would be helpful to have a clear separation for these pieces.

### DIFF
--- a/docker/builder/builder.go
+++ b/docker/builder/builder.go
@@ -47,7 +47,7 @@ type DaemonBuilder struct {
 // Build implements Builder. It consumes the docker build API endpoint and sends
 // a tar of the specified service build context.
 func (d *DaemonBuilder) Build(ctx context.Context, imageName string) error {
-	buildCtx, err := createTar(d.ContextDirectory, d.Dockerfile)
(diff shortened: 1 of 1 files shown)


vdemeester:
Although I agree, `CreateTar` is a little bit `docker` dependent (we talk about `Dockerfile` and we use `docker/pkg` package :anchor:).

@joshwget making use of it in `rancher` right ? :angel:

joshwget:
`CreateTar` is dependent on Docker concepts, but it's not dependent on a Docker daemon. I think the `docker/pkg` usage is pretty generic as well.

We do use this in Rancher Compose, and I'll admit that our use case might not be very common. `CreateTar` is used to create a build context and then the actual build process is handled by Rancher. Building is an important part of the Docker Compose experience though, so I think it's worth keeping this use case open for other implementations such as this.

vdemeester:
@joshwget yes definitely :wink: Was just curious but overall this LGTM 🐮

PR #324, opened 2016-07-08
Add EmptyNetworks struct

joshwget:
Add an `EmptyNetworks` struct similar to `EmptyService`.

Didn't think of this earlier, but we can use `EmptyNetworks` to get around having to always check that `networks` is non-nil like in #321.

### DIFF
--- a/project/empty.go
+++ b/project/empty.go
@@ -10,7 +10,7 @@ import (
 
 // this ensures EmptyService implements Service
 // useful since it's easy to forget adding new functions to EmptyService
-var _ Service = &EmptyService{}
+var _ Service = (*EmptyService)(nil)
 
 // EmptyService is a struct that implements Service but does nothing.
 type EmptyService struct {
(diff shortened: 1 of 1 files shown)


project/empty.go
```diff
@@ -120,3 +120,20 @@ func (e *EmptyService) Config() *config.ServiceConfig {
 func (e *EmptyService) Name() string {
 	return ""
 }
+
+// this ensures EmptyNetworks implements Networks
+var _ Networks = &EmptyNetworks{}
```
vdemeester: Even better `var _ Networks = (*EmptyNetworks)(nil)` should work too :angel: .

joshwget: Cool! I'll update :smile:

vdemeester:
> Didn't think of this earlier, but we can use `EmptyNetworks` to get around having to always check that `networks` is non-nil like in #321.

I like that 👼. I'm just wating for the CI to trigger to say LGTM :smile_cat:

vdemeester:
LGTM 🐸

PR #331, opened 2016-07-11
Remove Container.name field…

### LINKED ISSUE #262: Bug in collectContainers()
The code in service.go (collectContainers) does not work when Names has multiple strings of type /a/b and one or more of type /a. I ran into this.
Current code:
		// Compose add "/" before name, so Name[1] will store actaul name.
		name := strings.SplitAfter(container.Names[0], "/")
		result = append(result, NewContainer(client, name[1], containerNumber, s))

Fix:
                for i, val := range container.Names {
                  logrus.Debugf("Collect Container [%d]=%v", i, val)
                  name := strings.SplitAfter(val, "/")
                  if (name[0] == "/") && (len(name) == 2) {
                     result = append(result, NewContainer(client, name[1], containerNumber, s))
                  }
                }

vdemeester:
With the `Container`, it's not needed anymore 👼.

Fix #262 
Closes #234 
Closes #284

/cc @joshwget 

🐸

### DIFF
--- a/docker/container.go
+++ b/docker/container.go
@@ -29,7 +29,6 @@ import (
 type Container struct {
 	// FIXME(vdemeester) Replace with ContainerClient with engine-api vendor update
 	client    client.APIClient
-	name      string
 	id        string
 	container *types.ContainerJSON
 }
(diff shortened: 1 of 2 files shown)


docker/container.go
```diff
@@ -327,7 +324,7 @@ func (c *Container) Log(ctx context.Context, l logger.Logger, follow bool) error
 		Follow:     follow,
 		Tail:       "all",
 	}
-	responseBody, err := c.client.ContainerLogs(ctx, c.name, options)
+	responseBody, err := c.client.ContainerLogs(ctx, c.container.Name, options)
```
joshwget: Maybe `c.ID` is better here? From what I understand, we're trying to use IDs rather than names for API calls.

vdemeester: @joshwget oh you're right. I changed to `c.ID` in another PR in preparation, I should do it here :angel:. I'll update :wink:

joshwget:
LGTM

PR #338, opened 2016-07-21
Replace ListStoppedContainers by Containers

vdemeester:
This is more generic and allows us to add more filter in the future 👼.

I'm not sure the filter struct is the best way to do it, but that's the *first draft* 👼.

/cc @joshwget 


joshwget:
LGTM

vdemeester:
@joshwget I've added a `default` case to return an error if it's not a known `State` (https://github.com/docker/libcompose/pull/338/files#diff-f5e8a19e0479c40e7e311e1773a43a3bR527)

PR #339, opened 2016-07-26
 Add support to volume v2 🐡

### LINKED ISSUE #277: Composefile v2: support volumes

vdemeester:
This PR brings the first and main step to add support for volumes for v2 compose files 🐹. Thanks to @joshwget and @shouze 🐳

- Introduce new `yaml.Volume{,s}` struct
- Introduce new `docker/volume` package (like `network/volume`)

- [x] Make sure tests are a little bit complete (at least)

This closes #277

/cc @joshwget 

🐸

### DIFF
--- a/config/types.go
+++ b/config/types.go
@@ -116,7 +116,7 @@ type ServiceConfig struct {
 	ShmSize       yaml.StringorInt     `yaml:"shm_size,omitempty"`
 	StopSignal    string               `yaml:"stop_signal,omitempty"`
 	VolumeDriver  string               `yaml:"volume_driver,omitempty"`
-	Volumes       []string             `yaml:"volumes,omitempty"`
+	Volumes       *yaml.Volumes        `yaml:"volumes,omitempty"`
 	VolumesFrom   []string             `yaml:"volumes_from,omitempty"`
 	Uts           string               `yaml:"uts,omitempty"`
 	Restart       string               `yaml:"restart,omitempty"`
--- a/docker/convert.go
+++ b/docker/convert.go
@@ -16,6 +16,7 @@ import (
 	composeclient "github.com/docker/libcompose/docker/client"
 	"github.com/docker/libcompose/project"
 	"github.com/docker/libcompose/utils"
+	// "github.com/docker/libcompose/yaml"
 )
 
 // ConfigWrapper wraps Config, HostConfig and NetworkingConfig for a container.
@@ -34,6 +35,16 @@ func Filter(vs []string, f func(string) bool) []string {
 		}
 	}
 	return r
+}
+
+func toMap(vs []string) map[string]struct{} {
+	m := map[string]struct{}{}
+	for _, v := range vs {
+		if v != "" {
+			m[v] = struct{}{}
+		}
+	}
+	return m
 }
 
 func isBind(s string) bool {
@@ -58,21 +69,18 @@ func ConvertToAPI(serviceConfig *config.ServiceConfig, ctx project.Context, clie
 	return &result, nil
 }
 
-func isNamedVolume(volume string) bool {
-	return !strings.HasPrefix(volume, ".") && !strings.HasPrefix(volume, "/") && !strings.HasPrefix(volume, "~")
-}
-
-func volumes(c *config.ServiceConfig, ctx project.Context) map[string]struct{} {
-	volumes := make(map[string]struct{}, len(c.Volumes))
-	for k, v := range c.Volumes {
-		if len(ctx.ComposeFiles) > 0 && !isNamedVolume(v) {
-			v = ctx.ResourceLookup.ResolvePath(v, ctx.ComposeFiles[0])
+func volumes(c *config.ServiceConfig, ctx project.Context) []string {
+	if c.Volumes == nil {
+		return []string{}
+	}
+	volumes := make([]string, len(c.Volumes.Volumes))
+	for _, v := range c.Volumes.Volumes {
+		vol := v
+		if len(ctx.ComposeFiles) > 0 && !project.IsNamedVolume(v.Source) {
+			sourceVol := ctx.ResourceLookup.ResolvePath(v.String(), ctx.ComposeFiles[0])
+			vol.Source = strings.SplitN(sourceVol, ":", 2)[0]
 		}
-
-		c.Volumes[k] = v
-		if isVolume(v) {
-			volumes[v] = struct{}{}
-		}
+		volumes = append(volumes, vol.String())
 	}
 	return volumes
 }
@@ -141,6 +149,8 @@ func Convert(c *config.ServiceConfig, ctx project.Context, clientFactory compose
 		}
 	}
 
+	vols := volumes(c, ctx)
+
 	config := &container.Config{
 		Entrypoint:   strslice.StrSlice(utils.CopySlice(c.Entrypoint)),
 		Hostname:     c.Hostname,
@@ -154,7 +164,7 @@ func Convert(c *config.ServiceConfig, ctx project.Context, clientFactory compose
 		Tty:          c.Tty,
 		OpenStdin:    c.StdinOpen,
 		WorkingDir:   c.WorkingDir,
-		Volumes:      volumes(c, ctx),
+		Volumes:      toMap(Filter(vols, isVolume)),
 		MacAddress:   c.MacAddress,
 	}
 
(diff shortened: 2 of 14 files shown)


docker/convert.go
```diff
@@ -58,20 +58,19 @@ func ConvertToAPI(serviceConfig *config.ServiceConfig, ctx project.Context, clie
...
+		vol := v.Source
+		if len(ctx.ComposeFiles) > 0 && !project.IsNamedVolume(vol) {
+			vol = ctx.ResourceLookup.ResolvePath(vol, ctx.ComposeFiles[0])
+			v.Source = vol
 		}
 
-		c.Volumes[k] = v
-		if isVolume(v) {
-			volumes[v] = struct{}{}
+		// FIXME(vdemeester) why do we do that ?
```
vdemeester: I should remove that, this is handled [here](https://github.com/docker/libcompose/pull/339/files#diff-33042a808eb9153e032b7eb06ed82b0eR67) 👼

integration/volume_test.go
```diff
@@ -75,3 +75,23 @@ func (s *CliSuite) TestNamedVolume(c *C) {
 	c.Assert(cn.Mounts[0].Name, DeepEquals, "vol")
 	c.Assert(cn.Mounts[0].Destination, DeepEquals, "/path")
 }
+
+func (s *CliSuite) TestV2Volume(c *C) {
```
vdemeester: I shoud complete this one to make sure `Bind` and/or `Mounts` uses `project_test` as the name for the volume 👼

project/project.go
```diff
@@ -205,6 +206,32 @@ func (p *Project) load(file string, bytes []byte) error {
 	}
 
 	// Update network configuration a little bit
+	p.handleNetworkConfig()
+	p.handleVolumeConfig()
+
+	// FIXME(vdemeester) Not sure about this..
```
vdemeester: I should remove or rewrite this comment :joy:

project/project.go
```diff
@@ -238,21 +265,36 @@ func (p *Project) load(file string, bytes []byte) error {
...
+	return p.configVersion == "2"
+}
 
-		p.networks = networks
+func (p *Project) handleVolumeConfig() {
+	if p.isVolumeEnabled() {
+		for _, serviceName := range p.ServiceConfigs.Keys() {
+			serviceConfig, _ := p.ServiceConfigs.Get(serviceName)
+			// Consolidate the name of the volume
+			// FIXME(vdemeester) probably shouldn't be there, maybe move that to interface/factory
```
joshwget: It would definitely be useful if this logic could be customized :smile:

project/project.go
```diff
@@ -238,21 +265,36 @@ func (p *Project) load(file string, bytes []byte) error {
...
-		networks, err := p.context.NetworksFactory.Create(p.Name, p.NetworkConfigs, p.ServiceConfigs, p.isNetworkEnabled())
-		if err != nil {
-			return err
-		}
+func (p *Project) isNetworkEnabled() bool {
+	return p.configVersion == "2"
+}
 
-		p.networks = networks
+func (p *Project) handleVolumeConfig() {
```
joshwget: Not a huge deal, but can some of the nesting be removed in this function? It'd be a little easier to read.

config/types.go
```diff
@@ -116,7 +116,7 @@ type ServiceConfig struct {
 	ShmSize       int64                `yaml:"shm_size,omitempty"`
 	StopSignal    string               `yaml:"stop_signal,omitempty"`
 	VolumeDriver  string               `yaml:"volume_driver,omitempty"`
-	Volumes       []string             `yaml:"volumes,omitempty"`
+	Volumes       *yaml.Volumes        `yaml:"volumes,omitempty"`
```
joshwget: Should `ServiceConfigV1` have this change as well?

docker/volume/volume.go
```diff
@@ -0,0 +1,154 @@
...
+	projectName   string
+	name          string
+	driver        string
+	driverOptions map[string]string
+	external      bool
+	// TODO (shouze) missing labels
+}
+
+func (v *Volume) fullName() string {
+	name := v.projectName + "_" + v.name
```
joshwget: This logic seems to be duplicated in `handleVolumeConfig`. When we make it extensible there we should use the same thing here.

docker/volume/volume.go
```diff
@@ -0,0 +1,154 @@
...
+
+// Inspect inspect the current volume
+func (v *Volume) Inspect(ctx context.Context) (types.Volume, error) {
+	return v.client.VolumeInspect(ctx, v.fullName())
+}
+
+// Remove removes the current volume (from docker engine)
+func (v *Volume) Remove(ctx context.Context) error {
+	if v.external {
+		fmt.Printf("Volume %s is external, skipping", v.fullName())
```
joshwget: Would it be better to use `logrus.Infof` for these?

integration/volume_test.go
joshwget: I think an integration test for external volumes would be pretty useful as well.

docker/convert.go
```diff
@@ -58,21 +69,18 @@ func ConvertToAPI(serviceConfig *config.ServiceConfig, ctx project.Context, clie
...
-	volumes := make(map[string]struct{}, len(c.Volumes))
-	for k, v := range c.Volumes {
-		if len(ctx.ComposeFiles) > 0 && !isNamedVolume(v) {
-			v = ctx.ResourceLookup.ResolvePath(v, ctx.ComposeFiles[0])
-		}
-
-		c.Volumes[k] = v
-		if isVolume(v) {
-			volumes[v] = struct{}{}
+func volumes(c *config.ServiceConfig, ctx project.Context) []string {
```
vdemeester: This function do not update serviceConfig anymore 👼

docker/convert.go
```diff
@@ -36,6 +37,16 @@ func Filter(vs []string, f func(string) bool) []string {
 	return r
 }
 
+func toMap(vs []string) map[string]struct{} {
```
vdemeester: It's way to generic to not exists somewhere else, just didn't look at it for now 👼

vdemeester:
Unit tests / CI should be fixed 👼

docker/volume/volume.go
vdemeester: I did the same for `Network` for now :stuck_out_tongue:. I would like to keep it that way for 0.3.0 (to be coherent with networks), but in the end, this should be passed to a logger (not logrus, a libcompose logger) yes !

config/types.go
vdemeester: hum good question, but I don't think so, it will be converted to v2 anyway and it's this struct that is used everywhere else so, I don't see why we would want to update v1 structs

project/project.go
vdemeester: yes, same comment here and for network, this logic will have to be extracted at some point to be easier to customize :stuck_out_tongue_closed_eyes:

project/project.go
```diff
@@ -337,6 +383,16 @@ func (p *Project) Down(ctx context.Context, opts options.Down, services ...strin
 		return err
 	}
 
+	if opts.RemoveVolume {
```
joshwget: Should this be an option for `Delete` as well? I think the `rm` command also has a `-v` flag.

vdemeester: hum yes, `rm` has the `-v` flag :stuck_out_tongue:

project/project.go
vdemeester: I'm gonna update it but the way I reduce the nesting/complexity doesn't feel that much better :sweat_smile:

docker/volume/volume.go
vdemeester: Yeah I did the same for networks. I'll think about it and probably do some follow-up later about that. It's tricky because the 2 objects manipulated here and in `handleVolumeConfig` are different (one is a service volume, the other one is the declaration of the volume).

project/project.go
vdemeester: Ah, it's already handled for `rm` (but I'll have to check on `docker-compose` to see if it's removing the named volume too)

vdemeester: Ok, checked, `rm -v` does not remove declared/named volumes, so it's working the same :)

vdemeester:
As it's green, let's merge this and push the 0.3.0 milestone. Will work on coverage and refactoring in follow-ups.

Issue #336, opened 2016-07-18
Switch back to go-yaml


joshwget:
go-yaml now uses an Apache license (go-yaml/yaml#160) so the concerns in #79 should be addressed. There have been plenty of issues with candiedyaml, so it's probably best to switch back to go-yaml.

PR #340, opened 2016-07-26
Switch back to go-yaml

joshwget:
#336


vdemeester:
LGTM 🐹
And it's green so, *here we go* !

Issue #207, opened 2016-04-16
Complete up command


vdemeester:
There is still some flags/options to support for the `up` command : `--no-color`, `--no-deps`, `--build`, `--abort-on-exit` (see #162), `--timeout` and `--remove-orphans` .

```

Usage: up [options] [SERVICE...]

Options:
    -d                         Detached mode: Run containers in the background,
                               print new container names.
                               Incompatible with --abort-on-container-exit.
    --no-color                 Produce monochrome output.
    --no-deps                  Don't start linked services.
    --force-recreate           Recreate containers even if their configuration
                               and image haven't changed.
                               Incompatible with --no-recreate.
    --no-recreate              If containers already exist, don't recreate them.
                               Incompatible with --force-recreate.
    --no-build                 Don't build an image, even if it's missing.
    --build                    Build images before starting containers.
    --abort-on-container-exit  Stops all containers if any container was stopped.
                               Incompatible with -d.
    -t, --timeout TIMEOUT      Use this timeout in seconds for container shutdown
                               when attached or when containers are already
                               running. (default: 10)
    --remove-orphans           Remove containers for services not defined in
                               the Compose file
```

vdemeester:
- `--build`/`--no-build` are simple to implement, `no-color` and `--timeout` probably too
- `--remove-orhpan` and `--no-deps` a little bit more trick
- `--abort-on-container-ext` is gonna be harder (and depends/is related on/to https://github.com/docker/libcompose/pull/290)

kunalkushwaha:
> --build/--no-build are simple to implement, no-color and --timeout probably too
> --remove-orhpan and --no-deps a little bit more trick

I will try to complete these two first then

rdlf0:
Is `detached (-d)` option currently available and is it possible to pass a service name?

PR #342, opened 2016-07-27
--build option for up command

kunalkushwaha:
Added --build option for `up` command
Partial fix for #207

Signed-off-by: Kunal Kushwaha <kushwaha_kunal_v7@lab.ntt.co.jp>

### DIFF
--- a/docker/service.go
+++ b/docker/service.go
@@ -73,7 +73,7 @@ func (s *Service) Create(ctx context.Context, options options.Create) error {
 		return err
 	}
 
-	if err := s.ensureImageExists(ctx, options.NoBuild); err != nil {
+	if err := s.ensureImageExists(ctx, options.NoBuild, options.ForceBuild); err != nil {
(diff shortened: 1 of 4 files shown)


joshwget:
LGTM

docker/service.go
```diff
@@ -132,7 +132,11 @@ func (s *Service) collectContainers(ctx context.Context) ([]*Container, error) {
 	return result, nil
 }
 
-func (s *Service) ensureImageExists(ctx context.Context, noBuild bool) error {
+func (s *Service) ensureImageExists(ctx context.Context, noBuild bool, forceBuild bool) error {
```
vdemeester: We might want to pass the `options` here instead of boolean arguments :angel:

vdemeester: But that can be done in a follow-up though :angel:

vdemeester:
LGTM 🐸

PR #343, opened 2016-07-27
Parse multifile COMPOSE_FILE envvar, w/ side effect

### LINKED ISSUE #212: COMPOSE_FILE environment variable can contains multiple files
> The COMPOSE_FILE environment variable can now contain multiple files,
separated by the host system's standard path separator (: on Mac/Linux,
; on Windows).

Right now, I think `libcompose` assumes the `COMPOSE_FILE` environment file is only refering to one file… We should support the same as `docker-compose` does :angel:

dustinrc:
Fixes #212

urfave/cli does not distinguish whether the first string in the slice
comes from the envvar or is from a flag. Worse off, it appends the flag
values to the envvar value instead of overriding it. To ensure the
multifile envvar case is always handled, the first string must always be
split. It gives a more consistent behavior, then, to split each string
in the slice. With this side effect, the following is possible and
would merge the compose files left to right:

```
$ COMPOSE_FILE=a.yml:b.yml ./libcompose-cli -f c.yml -f d.yml:e.yml up
```

Signed-off-by: Dustin Chapman <dustin.r.chapman@gmail.com>

### DIFF
--- a/cli/command/command.go
+++ b/cli/command/command.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"os"
+	"strings"
 
 	"github.com/docker/libcompose/cli/app"
 	"github.com/docker/libcompose/project"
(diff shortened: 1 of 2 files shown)


bfosberry:
Nice!

vdemeester:
LGTM :cat:

PR #344, opened 2016-07-29
Add Remove command to ServiceConfigs

joshwget:

### DIFF
--- a/config/types.go
+++ b/config/types.go
@@ -201,6 +201,13 @@ func (c *ServiceConfigs) Add(name string, service *ServiceConfig) {
 	c.mu.Unlock()
 }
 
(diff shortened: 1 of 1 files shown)


vdemeester:
@joshwget any use case ? 👼

joshwget:
There's no way to "clear" a service config from a project yet. If a service config already exists for a project, `Project.Load` will merge the new config over the existing one. Together these two things mean you can't replace an existing config for a project, which is a pretty useful operation.

vdemeester:
@joshwget thx :angel: 
I'll add this to 0.3.0 as it's no big changes :stuck_out_tongue_closed_eyes: 

LGTM :cat:

PR #345, opened 2016-08-01
Extract `docker` code into smaller packages

vdemeester:
This extracts some code inside the `docker` package into smaller package, making it easier to isolate, reuse and tests 👼.

This defines 3 new packages : `auth`, `image` and `container`.

The `auth` package defines a way to lookup for Auth information (way to
log in to registries) and provide an implementation based on the
configfile implement in docker/docker.

The `image` package is responsible to handle operation on images (inspect,
pull, push, …).

The `container` package is responsible to hande operation on container (and defines mostly the `Container` struct and its methods).

Overall, to keep changes easier to review, this is merely a move refactoring, there is still TODOs and refactoring to do on these packages (few ideas are written in commit message).

/cc @joshwget 

*I would prefer this to be merged after the 0.3.0 release (which now only depends on #339)*

🐸


thaJeztah:
nice!

docker/image/image.go
```diff
@@ -16,20 +16,27 @@ import (
 	"github.com/docker/docker/registry"
 	"github.com/docker/engine-api/client"
 	"github.com/docker/engine-api/types"
+	"github.com/docker/libcompose/docker/auth"
 )
 
-func inspectImage(ctx context.Context, client client.APIClient, image string) (types.ImageInspect, error) {
+// InspectImage inspect the specified image (can be a name, an id or a digest)
+// with the specified client.
+func InspectImage(ctx context.Context, client client.ImageAPIClient, image string) (types.ImageInspect, error) {
```
vdemeester: Those function might be rename `Inspect`, `Remove` and `Pull` though (making it using like `image.Inspect(…)`, …

docker/service.go
```diff
@@ -460,31 +462,33 @@ func (s *Service) recreateIfNeeded(ctx context.Context, c *Container, noRecreate
 	return c, err
 }
 
-func (s *Service) recreate(ctx context.Context, c *Container) (*Container, error) {
+func (s *Service) recreate(ctx context.Context, c *container.Container) (*container.Container, error) {
 	name := c.Name()
-	newName := fmt.Sprintf("%s_%s", name, c.container.ID[:12])
+	id, _ := c.ID()
```
joshwget: Can `ID` be changed to not return an error? It seems to always be `nil` anyways.

vdemeester: Yes, I'll do that in a follow-up (just did the *fix* to make the move with the smallest changes possible)

joshwget:
LGTM!

PR #164, opened 2016-03-01
WIP - Adding Failure events to event types and to service wrapper

bfosberry:
This change adds a set of project and service level failure events to the
list of possible events, as well as to the service wrapper function calls.

This allows users to determine whether an action has failed by viewing events.

The only behaviour changed here, other than additional events passing through
the notification system, is that when a project action fails, it will no longer
post the relevant 'done' event.

PR #347, opened 2016-08-03
Adding strictly typed Events to libcompose

bfosberry:
This change updates all project, service and container events to be heavily typed,
which allows event readers to expect specific fields rather than reading from an
attribute map

Replaces https://github.com/docker/libcompose/pull/164

This is ready for review.

The main point of this change is that instead of using an event code enum, this now uses strict struct types. This allows select type-casting on the event read side, and as a result users can expect specific structures with clear fields and functions instead of having to look up specific values which may or may not be in the map and may or may not be valid as blank.

The downside of this approach is that it is very explicit and a lot of structs are generated. There is also some field misuse (project events have an un-used serviceName field), however the alternative for this would be to have two sets of interfaces for generating events (factory and wrapper)

To fit with the project/project and docker/service way of creating events en-masse, this ships with EventFactory (creates on-demand events for a specified service name) and EventWrapper (creates started/created/done events depending on if the action succeeded). This allow the struct typing to fit better into the more meta-approach of bulk-service actions using wrappers.

### DIFF
--- a/project/listener.go
+++ b/project/listener.go
@@ -1,35 +1,10 @@
 package project
 
 import (
-	"bytes"
+	"encoding/json"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/docker/libcompose/project/events"
-)
-
-var (
-	infoEvents = map[events.EventType]bool{
-		events.ServiceDeleteStart:  true,
-		events.ServiceDelete:       true,
-		events.ServiceDownStart:    true,
-		events.ServiceDown:         true,
-		events.ServiceStopStart:    true,
-		events.ServiceStop:         true,
-		events.ServiceKillStart:    true,
-		events.ServiceKill:         true,
-		events.ServiceCreateStart:  true,
-		events.ServiceCreate:       true,
-		events.ServiceStartStart:   true,
-		events.ServiceStart:        true,
-		events.ServiceRestartStart: true,
-		events.ServiceRestart:      true,
-		events.ServiceUpStart:      true,
-		events.ServiceUp:           true,
-		events.ServicePauseStart:   true,
-		events.ServicePause:        true,
-		events.ServiceUnpauseStart: true,
-		events.ServiceUnpause:      true,
-	}
 )
 
 type defaultListener struct {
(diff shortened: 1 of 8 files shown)


bfosberry:
Fixed a merge conflict

bfosberry:
I'll wait till we have possible changes to this branch or are about to merge to fix the next merge conflict

joshwget:
My only worry here is the maintenance for this, as you kind of mentioned. If a command is added, for example, there are many places where we'd have to update the events around it. Even before this PR we've had problems where we've forgotten to add the corresponding events. If you're willing to maintain this code in the future (whether or not that ends up requiring code generation), then I'm in favor of this.

bfosberry:
I think we can solve a lot of this with go generate and/or protobuf. The static compilation should ensure that a) automated processes using the lib can expect specific fields to exist in the events and b) that migrating to the changes will not leave code in an inconsistent state. I'd be happy to maintain this part of the codebase, I dont see myself working on projects that dont use this lib for the forseeable future. I think I need to work on this PR a little more to get it in a better state, but the rebasing and the major changes are difficult. 

I'll prototype this a little to try and and come up with a good generate pattern to cut down on the manual maintenance and difficult of new message support. I think there is also some room to refactor how we call and pass events around.

bfosberry:
Hey, so after some thoughts and research here is what I think we should do

1) Move even enums to be event json descriptions:

```
{
  "container_started": {
    "service_name": "abc",
    "container_id": "123",
  },
...
}
```

2) Use https://github.com/ChimeraCoder/gojson as a commit level tool to generate structs

At this stage all committed code and builds should resolve conflicts instead of having issues at runtime with map entries being missing or renamed.

3) Revisit generators and factories, will need to use go generate or similar to rebuild those dynamically

4) Strip service name out of most events since its only used on service level events, not project level.

Let me know what you think

joshwget:
Yeah, I think this will definitely look good after those changes!

project/listener.go
```diff
@@ -50,32 +25,93 @@ func NewDefaultListener(p *Project) chan<- events.Event {
...
+		} else if event.Service() == "" {
+			logf("Project [%s]: %s %s", d.project.Name, event.String(), data)
 		} else {
-			logf("[%d/%d] [%s]: %s %s", d.upCount, d.project.ServiceConfigs.Len(), event.ServiceName, event.EventType, buffer.Bytes())
+			logf("[%d/%d] [%s]: %s %s", d.upCount, d.project.ServiceConfigs.Len(), event.Service(), event.String(), data)
 		}
 	}
 }
+
+func infoLevel(event events.Event) bool {
```
GeorgeMac: Is this something you could build in to the type? Or is it dependent on the scope of receiving the message? e.g. an `*events.ServiceDeleteStart` is only info level in the project listener, but debug elsewhere? If log level is responsible to the event, perhaps then you could build level in to the events type. e.g. `event.LogLevel()` ? Just a thought :)

PR #353, opened 2016-08-09
Split project/project.go into smaller files

vdemeester:
Splits `project/project.go` into smaller pieces. To make it easier to navigate 👼

🐸


joshwget:
I'm impartial on this. It seems a little inconvenient if you have to make a change to each command, which is the case fairly often. Maybe all of the commands in one file and the rest of `project.go` in another would be something to consider.

If you'd prefer it this way and think others might as well, then LGTM.

vdemeester:
@joshwget well, `project/project.go` is not about commands — just about `Project` methods.
I'm *meh* too but I also don't like big files and it makes it easy to navigate :angel:. I'm also not sure we make that many changes to all methods at once :stuck_out_tongue_closed_eyes:

joshwget:
Okay, I'm pretty impartial on this so let's go ahead and merge.

PR #354, opened 2016-08-10
Simplify Info (as a map)

vdemeester:
`Info`/`InfoPart` does not really add anything more than a basic `map`, let's simplify that 👼

🐸


joshwget:
LGTM

PR #355, opened 2016-08-10
Move Service and Context to their own package.

vdemeester:
Continuing to split the `docker` package in several smaller packages, this time `service` and `ctx` 🐮.

- `service` package holds service related struct and methods. It's just
  a move, stuff like `Namer` should be moved away from this package and
  refactor in a way they don't call the docker API (e.g).
- `ctx` holds the `Context` struct used to pass factory and stuff :)

/cc @joshwget

🐸


project/project.go
```diff
@@ -15,6 +15,9 @@ import (
 	"github.com/docker/libcompose/yaml"
 )
 
+// ComposeVersion is name of docker-compose.yml file syntax supported version
+const ComposeVersion = "1.5.0"
```
vdemeester: We might want to *up* that sometimes as we support compose file v2 which is... way older than `1.5.0`👼

joshwget: If we switch this to "2", does that just indicate that we have support for v2 files? If so, I think it shouldn't hurt to bump this now.

vdemeester: it's based on the compose version (so something like `1.7.0`, …)

vdemeester: anyway, that's another *story* than this PR :stuck_out_tongue_closed_eyes:

joshwget:
LGTM

PR #359, opened 2016-08-11
Fix run command's service check logic

dustinrc:
The command would panic if no service were provided, or would not allow
running a service without additional comandParts.

Signed-off-by: Dustin Chapman <dustin.r.chapman@gmail.com>


vdemeester:
LGTM :cat:

PR #363, opened 2016-08-19
Clean up Makefile in libcompose

yuexiao-wang:
Clean up Makefile in libcompose and keep the same between rule and PHONY.

Signed-off-by: yuexiao-wang <wang.yuexiao@zte.com.cn>


vdemeester:
LGTM :cat:

PR #366, opened 2016-08-26
Rename script folder to hack

vdemeester:
It's mostly because `docker/docker` and most of the project I'm working
on uses `hack` instead of `script` 🐸

I'm so used to write `./hack/make.sh …`… each time I'm doing that here too…

/cc @joshwget 👼

🐸


joshwget:
If `hack` is pretty standard for Docker projects then I'm all for it. I'm used to `scripts` anyways :laughing:

vdemeester:
ah lol.. forgot to update the `Dockerfile`...

PR #367, opened 2016-08-26
Remove returned error in ID and IsRunning…

vdemeester:
… for containers. It wasn't needed anymore 🐯.

It also introduces a `container.ShortID()` method 👼.

/cc @joshwget 

🐸

Signed-off-by: Vincent Demeester vincent@sbr.pm

### DIFF
--- a/docker/container/container.go
+++ b/docker/container/container.go
@@ -168,9 +168,8 @@ func (c *Container) Kill(ctx context.Context, signal string) error {
 }
 
 // IsRunning returns the running state of the container.
-// FIXME(vdemeester): remove the nil error here
-func (c *Container) IsRunning(ctx context.Context) (bool, error) {
-	return c.container.State.Running, nil
+func (c *Container) IsRunning(ctx context.Context) bool {
+	return c.container.State.Running
 }
 
 // Run creates, start and attach to the container based on the image name,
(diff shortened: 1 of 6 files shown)


docker/container/container.go
```diff
@@ -355,6 +344,21 @@ func (c *Container) Networks() (map[string]*network.EndpointSettings, error) {
...
+	return c.container.ID
+}
+
+// ShortID return the container Id in its short form
+func (c *Container) ShortID() string {
+	return c.container.ID[:12]
+}
+
+// Name returns the container name.
+func (c *Container) Name() string {
```
joshwget: :+1:

joshwget:
Returning an error is useful for the interface. The `docker` package might not ever return an error, but other implementations might. I think it's particularly helpful for the use case where these functions aren't implemented ([example](https://github.com/rancher/rancher-compose/blob/master/rancher/container.go#L34)).

vdemeester:
@joshwget hum I get why it could be useful, I'm not sure how to do that though… :stuck_out_tongue_closed_eyes: (might be a *design* smell :sweat_smile:)

vdemeester:
@joshwget wdyt we should proceed on this ?

joshwget:
I guess this is okay since the `Containers` function returns an error in the interface (https://github.com/docker/libcompose/blob/master/project/project_containers.go#L140). It's still possible to return `ErrNotImplemented` there.

I'll try to do some thinking around how to handle the `Containers` interface and its functions in a better way. I always need to change things around a bit in our fork of libcompose, so I feel like the current abstractions could be improved.

vdemeester:
@joshwget so, should I merge this ? 👼

joshwget:
Sorry for the delay on this! LGTM

PR #371, opened 2016-08-31
Validation for v2 services

joshwget:
This handles all validation for v2 services except for service constraints, which is handled by #358. This does *not* handle validation for the `volumes` and `network` key.

To make this much easier to implement, I modified the v2 schema from Docker Compose so that the root is the `services` key. Most of the earlier error message generation makes this assumption and refactoring this would be a large effort.

I also had to remove one instance of `"additionalProperties": false` from the v2 schema since it seemed to cause an error in gojsonschema.

This does a little refactoring around how the schemas are loaded, but I think it can still be improved and custom schemas still cannot be used.

#357


config/schema.go
```diff
@@ -189,46 +190,14 @@ var schemaV1 = `{
 }
 `
 
-var schemaV2 = `{
+var schemaDataV2 = `{
```
vdemeester: If `service` is root here, we should name this variable accordingly I think 👼

joshwget: Updated!

vdemeester:
LGTM 🐸
Something weird on the CI, tested locally, it's green :)

PR #375, opened 2016-09-03
Use docker/ctx package in example in README

yudai:
docker.Context is not available any longer.

Signed-off-by: Iwasaki Yudai <yudai@arielworks.com>


vdemeester:
LGTM 🐸

Issue #377, opened 2016-09-04
Allow mutation of services after or during project creation


bfosberry:
I have a need to inject additional volumes and environment variables, as well as edit some port mappings depending on additional configuration parameters. 

My current method of achieving this is to cast APIProject -> Project and access p.ServiceConfigs, which is not ideal. My gut reaction is to add a "GetServiceConfig" function to the APIProject interface to allow this, but I'm not happy with exposing the config package out via the interface. Another option would be to have something like a `func MutateEnvironment(service string, func(currentEnv []string) []string)`, or maybe make a public service config interface for this kind of mutation. The ServiceConfig is basically just a struct with config params of base data types, so an interface might be overkill, maybe exposing it would be ok.

WDYT?

joshwget:
I'm not sure why exposing the `config` package would cause any problems. It sounds like the best solution here.

I've ran into a lot of similar limitations with `APIProject`. I've actually reverted back to using the `Project` struct in our fork of libcompose because it's easier to work with.

wagoodman:
This doesn't help with all possible changes to the config that you can make. For instance, if you want to modify a volume or add a service to a network, then changing the config would not help. For instance, when using Load() this would initialize the volume/network data structures, so changing the config at this point would do nothing (and is a bit misleading): https://github.com/docker/libcompose/blob/6bb86d53fa21f155e2daafa64796dbca61733df5/project/project.go#L210-L267

PR #379, opened 2016-09-06
Adding GetServiceConfig to project interface

bfosberry:
This change allows an APIProject type to return service configs for by name which
can be viewed and edited by the caller.

Solves #377

### DIFF
--- a/project/project.go
+++ b/project/project.go
@@ -98,7 +98,7 @@ func (p *Project) Parse() error {
 // CreateService creates a service with the specified name based. If there
 // is no config in the project for this service, it will return an error.
 func (p *Project) CreateService(name string) (Service, error) {
-	existing, ok := p.ServiceConfigs.Get(name)
+	existing, ok := p.GetServiceConfig(name)
 	if !ok {
 		return nil, fmt.Errorf("Failed to find service: %s", name)
 	}
@@ -502,6 +502,12 @@ func (p *Project) Notify(eventType events.EventType, serviceName string, data ma
 	for _, l := range p.listeners {
 		l <- event
 	}
+}
+
+// GetServiceConfig looks up a service config for a given service name, returning the ServiceConfig
+// object and a bool flag indicating whether it was found
+func (p *Project) GetServiceConfig(name string) (*config.ServiceConfig, bool) {
+	return p.ServiceConfigs.Get(name)
 }
 
 // IsNamedVolume returns whether the specified volume (string) is a named volume or not.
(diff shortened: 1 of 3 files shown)


vdemeester:
@mikedougherty the CI seems broken 👼

project/project.go
```diff
@@ -18,6 +18,11 @@ import (
 // ComposeVersion is name of docker-compose.yml file syntax supported version
 const ComposeVersion = "1.5.0"
 
+var (
+	// ErrNoConfig is returned when a service config cannot be found by name
+	ErrNoConfig = errors.New("Failed to find service")
```
vdemeester: I'm not found of exported errors as they couple the caller (or the code handling the error) to the library :stuck_out_tongue_closed_eyes: (cf http://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully)

vdemeester:
Overall I'm fine with that.
/cc @joshwget

bfosberry:
Great! I'll fix the error and get CI passing as best I can

mikedougherty:
@vdemeester thanks for the ping; looks like a recent plugin upgrade caused the run to hit a script access exception. I've fixed that issue and things look to be chugging along here: https://jenkins.dockerproject.org/job/docker/job/libcompose/job/PR-379/3/consoleFull

vdemeester:
@mikedougherty thanks 👼

mikedougherty:
@vdemeester fyi it looks like the result of the acceptance tests is ignored -- there seem to be a lot of test failures in the log but it still exits cleanly.

vdemeester:
@mikedougherty yep, this is the compose acceptance tests — it's here to give an idea of how close our cli is to the docker-compose one. In an ideal world we will not ignore it at some point, but there is still some ways to go (and some tests to think of as it's gonna be difficult to make them green)

:stuck_out_tongue_closed_eyes:

bfosberry:
I updated this to fix the error, I think this is gtg

vdemeester:
LGTM 🐸

joshwget:
Maybe we can just return a bool instead of an error?

bfosberry:
Hmm, I'm fine with that, I can move the error return up and maintain functionality

vdemeester:
Hum why a bool ? we hides potential error cause by doing that 
😅

bfosberry:
So from an external perspective using the api and getting a service by an invalid name may not be an error, it may may be more like a map lookup. nternally we still want to bubble up an error, but I think it makes sense as a bool for the project api return type.

joshwget:
Would you mind squashing @bfosberry? Otherwise LGTM!

bfosberry:
You got it

bfosberry:
gtg, thanks guys!

PR #383, opened 2016-09-13
Adding ServiceConfigs function to get dependent services

bfosberry:
This change adds in a function to config/ServiceConfigs to provide back a list of dependent services required to start a specified service, including the passed service. This is recursively generated based on the docker/compose functionality here: https://github.com/docker/compose/blob/acfe100686fd95d524ff102c0b5fccff0bc79d8c/compose/service.py#L519

This can be used to translate a service name into a service array which can be passed directly to a compose action.

e.g.

```
app:
  build: ./
  links:
   - db
db:
  image: postgres
test:
  build: ./
  dockerfile_path: Dockerfile.test
```

```
 services := serviceConfig.GetDependentServices("app") //returns ["app", "db"]
 err = project.Up(context.Background(), options.Up{}, services...)
```

In a later PR we can bake this change in by default and have actions automatically determine dependencies

### DIFF
--- a/config/types.go
+++ b/config/types.go
@@ -1,9 +1,14 @@
 package config
 
 import (
+	"strings"
 	"sync"
 
 	"github.com/docker/libcompose/yaml"
+)
+
+const (
+	servicePrefix = "service:"
 )
 
 // EnvironmentLookup defines methods to provides environment variable loading.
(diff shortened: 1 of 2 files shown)


joshwget:
Logic for handling dependencies already exists (https://github.com/docker/libcompose/blob/master/project/utils.go#L9). Is there a reason for implementing this again in the `config` package?

bfosberry:
I had asked around and hadn't been able to locate it. I can switch to use
that, but it does not work recursively it seems

bfosberry:
So we need a way to recursively form the dependency set, and we can also do this at the ServiceConfig level not the project level. @joshwget thoughts on changing the util function to work recursively and drop project for serviceconfig

bfosberry:
ping @joshwget what do you think of making these changes?

joshwget:
It's probably better to do this at the project level since dependencies can be extended via the `Service` interface (https://github.com/docker/libcompose/blob/master/project/service.go#L34). We'd miss out on [these dependencies](https://github.com/docker/libcompose/blob/master/docker/service/utils.go#L11), for example.

vdemeester:
@bfosberry the basic idea is to be able to start dependent services (recursively) right, and only those ? Like @joshwget I definetely think it should be done at the project level, maybe even a special *up* function or something ? :angel:

bfosberry:
So in order to maintain parity with docker/compose we need something in the main project functions that recursively resolves dependencies by default. I dont think it should be a special variant of up.

docker/compose get_services calls _inject_deps which calls get_services: https://github.com/docker/compose/blob/master/compose/project.py#L495

I can move the implementation in this PR over to the project level, and have it use the existing util in a recursive manner removing duplicates.

bfosberry:
@vdemeester yes btw the goal is to start only those services we need

joshwget:
@vdemeester What do you think about the approach of using a flag for `Up` where the default is to bring up dependencies?

vdemeester:
@joshwget @bfosberry seems good to me :wink:

bfosberry:
Great! Will sort it out

bfosberry:
So this is getting complicated. Basically docker/service/DependentServices calls project/DependentServices. So I think it makes more sense to have this at the service level

bfosberry:
Closing this branch out in favour of a fresh implementation

Issue #381, opened 2016-09-09
No way to specify empty environment variables


joshwget:
There isn't currently a way to specify empty environment variables for a service. For the following template, I'd expect `TEST` to appear as an empty environment variable.

```
test:
  image: ubuntu
  command: env
  environment:
    TEST: ""
```
```
$ ./bundles/libcompose-cli up                   
test_1 | PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
test_1 | HOSTNAME=c5f0ab2fdd94
test_1 | HOME=/root
```

vdemeester:
@joshwget sounds fair, `docker/compose` creates an empty enviornment variable so this should be considered as a bug.

joshwget:
Fixed by #384

PR #384, opened 2016-09-16
Fix empty environment variable

weitenghuang:
Fix issue [#381](https://github.com/docker/libcompose/issues/381)

Signed-off-by: weitenghuang <weiteng.huang@concur.com>

### DIFF
--- a/project/project.go
+++ b/project/project.go
@@ -134,7 +134,7 @@ func (p *Project) CreateService(name string) (Service, error) {
 
 		for _, env := range config.Environment {
 			parts := strings.SplitN(env, "=", 2)
-			if len(parts) > 1 && parts[1] != "" {
+			if len(parts) > 1 {
 				parsedEnv = append(parsedEnv, env)
 				continue
 			} else {
--- a/project/project_test.go
+++ b/project/project_test.go
@@ -153,7 +153,7 @@ func TestEnvironmentResolve(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if !reflect.DeepEqual(service.Config().Environment, yaml.MaporEqualSlice{"A=X", "A=X", "A=B"}) {
+	if !reflect.DeepEqual(service.Config().Environment, yaml.MaporEqualSlice{"A=X", "A=", "A=B"}) {
 		t.Fatal("Invalid environment", service.Config().Environment)
 	}
 }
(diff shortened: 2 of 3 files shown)


joshwget:
This feels like the right fix, but at the same time I think the check was likely there for a reason. Since we don't have many tests around environment at the moment, we're at a pretty high risk for a regression if we merge this without tests. Since environment can be both a list and a map, there a lot of cases to consider.

Empty variable for list
```
environment:
- A=
```

Regular variable for list
```
environment:
- A=A
```

Lookup variable for list
```
environment:
- A
```

Empty variable for map
```
environment:
  A: ""
```

Regular variable for map
```
environment:
  A: A
```

Lookup variable for map
```
environment:
  A:
```

I'm happy to write these tests if you don't have time, but I think it would be good to have these tests for this change.

weitenghuang:
@joshwget good point. Let me add tests for it. 
Btw, when I run `docker-compose` for those cases, it doesn't seem to be exactly the same between `map` and `list`.
Did I misunderstand it?
For example:
```yaml
test:
  image: ubuntu
  command: env
  environment:
    TEST: ""
```
result in
```sh
test_1  | PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
test_1  | HOSTNAME=06a947584611
test_1  | TEST=
test_1  | no_proxy=*.local, 169.254/16
test_1  | HOME=/root
```
while 
```yaml
test:
  image: ubuntu
  command: env
  environment:
    - TEST=""
```
result in 
```sh
test_1  | PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
test_1  | HOSTNAME=e64b2e7ad46c
test_1  | TEST=""
test_1  | no_proxy=*.local, 169.254/16
test_1  | HOME=/root
```

My docker version:
```sh
libcompose fix-381 % docker-compose --version
docker-compose version 1.8.0-rc1, build 9bf6bc6
libcompose fix-381 % docker --version
Docker version 1.12.0-rc2, build 906eacd, experimental
```

joshwget:
I think that sounds correct to me. In the first case `""` is interpreted as an empty string in YAML. In the second case the value for the environment variable is set as the raw string `""`.

weitenghuang:
@joshwget , i've to update one existing test (since we consider the original logic was a bug) and also add new ones for those cases. Hope this is good to go now.

joshwget:
LGTM. Thanks @weitenghuang!

PR #389, opened 2016-09-22
Create config separately

mrajashree:
This PR separates the creation of config from `Merge` function and puts it in `CreateConfig`. This is because unmarshmaling of bytes to config or RawServiceMap is required by `Merge`, `MergeServicesV1`, `MergeServicesV2` and `ParseV2`, and is currently done in `Merge`. The same code of config creation from `Merge` can be used in other places, but `Merge` does many other things so it can't be called from all functions. These functions that require config creations can directly call the function `CreateConfig` for that.  Based on the config.Version, for version 2, `CreateConfig` checks that no values in `config.Networks` and `config.Volumes` are `nil`. So the functions `ParseNetworks` and `ParseVolumes` can be removed. If version is not 2, `CreateConfig` creates a `RawServiceMap` and assigns to `config.Services`.

### DIFF
--- a/config/merge.go
+++ b/config/merge.go
@@ -23,36 +23,55 @@ var (
 	}
 )
 
+// CreateConfig unmarshals bytes to config and creates config based on version
+func CreateConfig(bytes []byte) (*Config, error) {
+	var config Config
+	if err := yaml.Unmarshal(bytes, &config); err != nil {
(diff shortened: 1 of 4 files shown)


vdemeester:
Thanks @mrajashree :+1:

PR #391, opened 2016-09-30
WIP - Adding recursive dependency fetching

bfosberry:
Still need to patch in a flag to indicate recursive or not

replaces https://github.com/docker/libcompose/pull/383


bfosberry:
Hey @vdemeester, @joshwget, what do you guys think of extracting the docker/project from an action on it. Essentially make a wrapper to store a lot of that metadata around actions on the project like start/stop/done events, and whether to use recursive deps. Most of those fields seem unwieldly when passed around as args, and dont belong as fields on the project struct. 

WDYT?

PR #394, opened 2016-10-12
Add tmpfs

joshwget:


joshwget:
It should be caught by validation on the CLI side, so I'm thinking it might be better to leave it in case there's a library use case for it.

vdemeester:
@joshwget sounds good to me :stuck_out_tongue_closed_eyes: 
Needs a rebase though :angel:

PR #409, opened 2016-10-29
Add filtering on project ps

halfa:
Implements correct filtering for project services listing as specified in documentation.

### DIFF
--- a/project/project_ps.go
+++ b/project/project_ps.go
@@ -1,6 +1,8 @@
 package project
 
 import (
+	"sort"
+
 	"golang.org/x/net/context"
 )
 
@@ -8,6 +10,14 @@ import (
 func (p *Project) Ps(ctx context.Context, services ...string) (InfoSet, error) {
 	allInfo := InfoSet{}
 	for _, name := range p.ServiceConfigs.Keys() {
+		if services != nil { // apply filter
+			sort.Strings(services)
+			index := sort.SearchStrings(services, name)
+			if index > len(services) {
+				continue
+			}
+		}
+
 		service, err := p.CreateService(name)
 		if err != nil {
 			return nil, err
--- a/project/project_ps.go
+++ b/project/project_ps.go
@@ -1,28 +1,16 @@
 package project
 
-import (
-	"sort"
-
-	"golang.org/x/net/context"
-)
+import "golang.org/x/net/context"
 
 // Ps list containers for the specified services.
 func (p *Project) Ps(ctx context.Context, services ...string) (InfoSet, error) {
 	allInfo := InfoSet{}
 
-	if services != nil {
-		sort.Strings(services)
+	if services == nil {
+		services = p.ServiceConfigs.Keys()
 	}
 
-	for _, name := range p.ServiceConfigs.Keys() {
-		if services != nil { // apply filter
-			index := sort.SearchStrings(services, name)
-			// index hold the position where the data should be,
-			// be it present or not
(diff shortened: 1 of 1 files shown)


joshwget:
Can you explain a little more on what the goal is here? I'm not sure what about filtering is different between libcompose and Docker Compose.

halfa:
This isn't a difference between `libcompose` and Docker Compose.
This PR aims to fix the fact that the `services ...string` parameter is unused and the godoc read "for the specified services".

project/project_ps.go
```diff
@@ -1,13 +1,23 @@
...
+
 	"golang.org/x/net/context"
 )
 
 // Ps list containers for the specified services.
 func (p *Project) Ps(ctx context.Context, services ...string) (InfoSet, error) {
 	allInfo := InfoSet{}
 	for _, name := range p.ServiceConfigs.Keys() {
+		if services != nil { // apply filter
+			sort.Strings(services)
```
joshwget: This sort could be moved to outside the for loop, right? It might actually be better to just build up the list of needed services prior to this for loop.

halfa:
Actually I wrote this code quite a while ago... This version is much better. Should I write some testing on this one ?

joshwget:
I'm not actually sure why any sorting or searching through a list is necessary here. We can just iterate on `services` instead of `p.ServiceConfigs.Keys()`. If `services == nil`, then we can just set `services = p.ServiceConfigs.Keys()`.

halfa:
I wanted to avoid possibilities of a non-existent service being processed farther, sorting was to use binary search. If checking is out of scope I'll swap the `services` array. Working on the failing test.

joshwget:
I'm not too worried non-existent services. They'll just cause an error in `CreateService`.

project/project_ps.go
```diff
@@ -1,13 +1,17 @@
 package project
 
-import (
-	"golang.org/x/net/context"
-)
+import "golang.org/x/net/context"
```
vdemeester: I don't like that change :stuck_out_tongue_closed_eyes:

joshwget: Isn't this pretty standard convention? I think it might actually be `gofmt` that makes this change.

halfa: Yes that's a `gofmt` convention :grinning:

vdemeester: erf, I would be sad if `gofmt` did that :sweat_smile:

halfa:
I'm still puzzled about failing CI tests - actually `make test` fail even on master on my laptop, so I'll need some time to figure out why...

project/project_ps.go
halfa: Feel free to revert this chunk if you don't like it :')

vdemeester: ain't no big deal anyway :stuck_out_tongue_closed_eyes:

project/project_ps.go
```diff
@@ -1,13 +1,17 @@
...
 
 // Ps list containers for the specified services.
 func (p *Project) Ps(ctx context.Context, services ...string) (InfoSet, error) {
 	allInfo := InfoSet{}
-	for _, name := range p.ServiceConfigs.Keys() {
+
+	if services == nil {
```
vdemeester: Tests are correctly failing. This line should be `if len(services) == 0`. Otherwise it does not display any services from the CLI (because `services ...string` is not `nil`, it's an empty array).

Issue #399, opened 2016-10-13
example doesn't compile


geoffhendrey:
if I import this "github.com/docker/libcompose/docker/ctx"
and change the first line to this:
project, err := docker.NewProject(&ctx.Context{

then it works

joshwget:
Yeah, the example often gets overlooked. Want to put in a PR to fix it? :smile:

vdemeester:
Ohhhh… we did not revert the example :sweat:

PR #410, opened 2016-10-31
Fix example

kszarek:
Solves https://github.com/docker/libcompose/issues/399


joshwget:
LGTM. Thanks @kszarek!

PR #411, opened 2016-11-04
Supporting human readable strings for mem_limit

kohlerm:
Currently libcompose fails to parse compose files that use human readable strings for mem_limit, such as 40m for 40Mbyte. 
These simple changes would make it work, but would also have the side effect that libcompose would accept this kind of human readable string in other places as well (where it would not make sense).
It is therefore not a complete solution.  
I wonder how a complete solution would look like.Would we need to introduce another type
MemStringorInt, which would be used wherever memory values are specified?

### DIFF
--- a/project/project_test.go
+++ b/project/project_test.go
@@ -243,5 +243,5 @@ func TestParseWithMultipleComposeFiles(t *testing.T) {
 	assert.Equal(t, "busybox", multipleConfig.Image)
 	assert.Equal(t, "multi", multipleConfig.ContainerName)
 	assert.Equal(t, []string{"8000", "9000", "10000"}, multipleConfig.Ports)
-	assert.Equal(t, yaml.StringorInt(41943040), multipleConfig.MemLimit)
+	assert.Equal(t, yaml.MemStringorInt(41943040), multipleConfig.MemLimit)
 }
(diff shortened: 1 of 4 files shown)


joshwget:
I've been hoping something like `RAMInBytes` existed so we wouldn't have to write it. Thanks for finding that!

I definitely agree with your proposed solution. A separate type for memory strings makes sense to me.

vdemeester:
I think `docker/compose` support both notation right ? We should update the tests to make sure we do too 👼

joshwget:
Yeah, Docker Compose does support this. @kohlerm would you mind writing some tests for this? This would be great to merge!

kohlerm:
Do you have any preference of where that test should be located? I already slightly changed an existing test in this pull request. Admittedly that test alone does not doesn't provide a lot of test coverage.

kohlerm:
Ok, tests are green again. Sorry took a bit longer, because I could not yet get the integration test running locally (some http_proxy issues). 
Any recommendations for where additional tests should be added?

vdemeester:
@kohlerm could you squash your commits ? 👼

project/project_test.go
```diff
@@ -201,7 +201,7 @@ func TestParseWithMultipleComposeFiles(t *testing.T) {
 	configThree := []byte(`
```
vdemeester: Idea would be to add a `configFour` with the `mem_limit: 40000000` to make sure it still works.

PR #412, opened 2016-11-05
Refactor and expose convertKeysToStrings

joshwget:
`convertKeysToStrings` is a really useful function that I use across multiple projects. It's also pretty general, so it makes sense to move this to the `utils` package.

### DIFF
--- a/config/validation.go
+++ b/config/validation.go
@@ -5,6 +5,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/docker/libcompose/utils"
 	"github.com/xeipuuv/gojsonschema"
 )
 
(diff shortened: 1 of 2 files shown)


vdemeester:
It make sence to expose it and probably move into another package, I'm just not sure if `utils` is the right one (`utils` has no real meaning and is a "I don't know where to put this package" — I'm not found of it and tend to do all I can to not have a `utils` package 👼). Not sure what the name of the package should be though :sweat_smile:

joshwget:
I agree, there's a bit too much going on in `utils`. If you have any suggestions we can address it now. Or maybe it's better to move this discussion to a separate issue.

PR #417, opened 2016-12-01
Create the default network if required.

### LINKED ISSUE #319: networking config v2
Hey!

Is there something specific that I need to do to get networking config to work in v2?  Using docker-compose, the following config will result in a "default" network config labled after the project name to be connected to the container started up:

```
version: '2'
services:
  org-api.app.local:
    image: org-api:testbob
```

Is there something more than just:

`APIProject.Up(context.Background(), options.Up{options.Create{ForceRecreate: true}}, service_name)`

needed?  Do I need to specifically create the networks?  I did not see anything relevant to this in the CLI code, and while I see a number of functions related to networking, it's not clear to me what I need to do with them.

vdemeester:
This make sure we create and use a default network if no other networks exists and/or some services don't use a defined network.

- [ ] Needs more tests :angel: 
- [x] I think I need to cover the case where one service has a defined network but not the other one

Should fix #319 

/cc @joshwget @dansteen 

🐸

### DIFF
--- a/docker/network/network.go
+++ b/docker/network/network.go
@@ -169,6 +169,12 @@ func NetworksFromServices(cli client.NetworkAPIClient, projectName string, netwo
 			networkNames[network.Name] = network
 		}
 	}
+	if len(networkConfigs) == 0 {
+		network := NewNetwork(projectName, "default", &config.NetworkConfig{
+			Driver: "bridge",
+		}, cli)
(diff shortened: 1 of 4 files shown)


vdemeester:
@lussoluca good. There is still a case where it might not create it… I'll have to update this (or follow-up)

PR #422, opened 2016-12-27
Carry #409 — Add filtering on project ps

vdemeester:
Carry #409 – Implements correct filtering for project services listing as specified in documentation.

Closes #409 

👼 🐸


halfa:
Good job ;)

PR #432, opened 2017-02-02
Use vndr to vendor stuff

vdemeester:
Removing shell scripts. I'll update the image to put `vndr` inside so it can be used for update vendoring 👼.


joshwget:
+1 for moving to vendor.conf. Any change you'd consider trash though? I reference the imports for libcompose very often for other projects and it would be nice to know the vendor folders are going to be the same.

Hopefully trash and vndr will produce the same vendor folders soon, but that's another conversation :smile:

vdemeester:
@joshwget I used `vndr` because this is the one we tend to use for projects at Docker 👼. 

>  I reference the imports for libcompose very often for other projects and it would be nice to know the vendor folders are going to be the same.
> Hopefully trash and vndr will produce the same vendor folders soon, but that's another conversation 😄

What would differ 😛 ? I mean any diff in non-go files is ok I would think. Other than that, as long as both remove `_test.go` and non used package, I wonder what would differ, or more importantly how would it affect the downstream project ? 👼

joshwget:
Yeah, I guess I don't really have a good case here other than being pretty familiar with trash already. Looks like https://github.com/golang/dep is the future anyways.

LGTM once green!

Issue #435, opened 2017-02-06
libcompose fails when float environment value is passed 


surajnarwade:
when we pass float value in environment variables as below:
```  
environment:
    TEST: 0.2
 ```  
I am getting following error:
```
ERRO[0000] Could not parse config for project compose : Cannot unmarshal '0.2' of type float64 into a string value 
FATA[0000] Failed to read project: Cannot unmarshal '0.2' of type float64 into a string value
```

PR #437, opened 2017-02-08
Allows the use of float values in environment variables

cdrage:
When converting a docker-compose file with float values as environment
variables, libcompose errors out.

Example:

```
version: "2"

services:
    redis:
      image: redis:3.0
      environment:
        foo: 0.3
```

Errors out with: `Cannot unmarshal '0.3' of type float64 into a string value`

This commit fixes the issue by detecting if a float value has been
passed and successfully converting it to a string variable.


cdrage:
Closes #435

cdrage:
ping @joshwget @surajssd @snarwade

surajnarwade:
LGTM :+1: , it works for me :)

Issue #443, opened 2017-02-27
.env file lookup in current directory rather than docker-compose file directory


surajnarwade:
when I do from same directory where docker-compose file is placed, everything works fine
```
$ libcompose up
```
but, when I do from any other directory, like 
```
libcompose -f test/docker-compose.yml up
```
it fails, it looks for .env file in current directory rather than test.

vdemeester:
For reference : https://github.com/docker/docker.github.io/blob/master/compose/compose-file/index.md#env_file 👼

surajnarwade:
I want to work on it

surajnarwade:
cc @vdemeester

PR #448, opened 2017-03-03
Fixes .env file lookup

surajnarwade:
fixes issue #443

### DIFF
--- a/project/project.go
+++ b/project/project.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -65,7 +66,17 @@ func NewProject(context *Context, runtime RuntimeProject, parseOptions *config.P
 	}
 
 	if context.EnvironmentLookup == nil {
-		cwd, err := os.Getwd()
+		var envPath, absPath, cwd string
+		var err error
+		if len(context.ComposeFiles) > 0 {
+			absPath, err = filepath.Abs(context.ComposeFiles[0])
+			dir, _ := path.Split(absPath)
+			envPath = filepath.Join(dir, ".env")
+		} else {
+			cwd, err = os.Getwd()
+			envPath = filepath.Join(cwd, ".env")
+		}
+
 		if err != nil {
 			log.Errorf("Could not get the rooted path name to the current directory: %v", err)
 			return nil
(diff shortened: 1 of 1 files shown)


surajnarwade:
cc @vdemeester  @joshwget

project/project.go
```diff
@@ -66,14 +67,30 @@ func NewProject(context *Context, runtime RuntimeProject, parseOptions *config.P
 
 	if context.EnvironmentLookup == nil {
 		cwd, err := os.Getwd()
+		var isPresent bool
```
joshwget: You should be able to read `ComposeFiles` from context rather than looking through `os.Args`. You can use the directory of the first Compose file.

project/project.go
```diff
@@ -65,15 +65,23 @@ func NewProject(context *Context, runtime RuntimeProject, parseOptions *config.P
 	}
 
 	if context.EnvironmentLookup == nil {
-		cwd, err := os.Getwd()
+		var envPath, absPath, dir string
+		var err error
+		for _, file := range context.ComposeFiles {
```
joshwget: We should just look at the first Compose file. This will always look at the last.

project/project.go
```diff
@@ -65,15 +65,18 @@ func NewProject(context *Context, runtime RuntimeProject, parseOptions *config.P
 	}
 
 	if context.EnvironmentLookup == nil {
-		cwd, err := os.Getwd()
+		absPath, err := filepath.Abs(context.ComposeFiles[0])
```
joshwget: Can you check to make sure that `len(context.ComposeFiles[0]) > 0`? If not we should fall back to the existing behavior (`Path: filepath.Join(cwd, ".env")`).

surajnarwade: no need of that, this code covers all conditions

joshwget: The `app` package guarantees that `len(context.ComposeFiles) > 0`, but it's possible to create a project using `ComposeBytes` rather than `ComposeFiles`.

surajnarwade:
cc @joshwget

joshwget:
LGTM

project/project.go
```diff
@@ -16,6 +15,8 @@ import (
 	"github.com/docker/libcompose/project/events"
 	"github.com/docker/libcompose/utils"
 	"github.com/docker/libcompose/yaml"
+	"os"
```
vdemeester: It's a nit but, these imports should be on top (with other built-in ones). It's weird gofmt didn't fail on that.

surajnarwade:
@vdemeester done :)

PR #469, opened 2017-05-18
Fix GetServiceHash failed due to false hashing

Azure22:
Hashing *yaml.Networks and *yaml.Volumes in previous code was processed wrong. Sprintf an array of pointers by %v in default will get a string of mem addresses, which could be changing every time libcompose project up the same compose file. Therefore OutOfSync was always triggered if service had a network config.


Azure22:
Hi. Should I close this pr and sign the commit and make a new one? Very first time here :)

vdemeester:
@Azure22 you should sign your commit and force push your branch :wink: no need to re-open a new one :angel:

Azure22:
@GordonTheTurtle @vdemeester Thanks for your detailed reply!!

PR #492, opened 2017-09-08
adding ability to parse default bash variables such as ${foo:-bar} 

vito-c:
title is pretty straight forward. This let's you add default values to your docker-compose file.
```
web:
# unbracketed name
image: ${IMAGE:-busybox}
```
parsers to
```
web:
# unbracketed name
image: busybox
```
```
web:
# unbracketed name
image: ${IMAGE-busybox}
```
parsers to
```
web:
# unbracketed name
image: busybox
```

### DIFF
--- a/config/interpolation.go
+++ b/config/interpolation.go
@@ -15,7 +15,7 @@ func isNum(c uint8) bool {
 }
 
 func validVariableDefault(c uint8, line string, pos int) bool {
-	return c == ':' && line[pos+1] == '-'
+	return (c == ':' && line[pos+1] == '-') || (c == '-')
 }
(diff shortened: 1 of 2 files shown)


patrick-bohan:
Does this support the `${VARIABLE-default}` format (without the ":") as well? https://docs.docker.com/compose/compose-file/#variable-substitution

vito-c:
@patrick-bohan no but I could add support for that.

vito-c:
added support for `-` and some more testing.

vito-c:
@vdemeester What is the timeline like to get merges?

PR #496, opened 2017-09-29
Make defaults work with nested env variables ie: ${FOO:-${HOME}/bar}

vito-c:
Signed-off-by: GodFather <vito.blog@gmail.com>


cheapRoc:
👍 It would be great to get this in since I've seen this fail on some of my compose files.

vito-c:
@vdemeester how's it looking on this one? :)

vito-c:
It feels like they actually have a bug in their code. The default string is not interpolated correctly. If you try something like this `echo ${ABC:-${HOME}}` you'll notice it will output your homedir.

```
---
version: '3'
services:
    test.env:
        container_name: test.env
        image: alpine
        environment:
            FOO: ${ABC:-foo}
        command: env
```
output:
```
Recreating test.env ...
Recreating test.env ... done
Attaching to test.env
test.env    | PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
test.env    | HOSTNAME=0e6eab0bcf66
test.env    | FOO=foo
test.env    | no_proxy=*.local, 169.254/16
test.env    | HOME=/root
test.env exited with code 0
```

and
```
---
version: '3'
services:
    test.env:
        container_name: test.env
        image: alpine
        environment:
            FOO: ${ABC:-${HOME}}
        command: env
```
output: 
```
Recreating test.env ...
Recreating test.env ... done
Attaching to test.env
test.env    | PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
test.env    | HOSTNAME=6fa2fe19685b
test.env    | FOO=${HOME}
test.env    | no_proxy=*.local, 169.254/16
test.env    | HOME=/root
test.env exited with code 0
```

vdemeester:
Yeah, that's why I would like to make this behavior optional (so ppl can opt-in or opt-out of the behavior to get the same as `docker-compose`.)

vito-c:
sweet how would I go about adding the optional flag?

vdemeester:
@vito-c I'm not sure yet. I would probably make `Interpolate` *configurable* (https://github.com/vito-c/libcompose/blob/628498254cf9da3ab00dad397b64a17d11d65140/config/interpolation.go#L184) that would be disabled by default (on `libcompose` binary).

Issue #441, opened 2017-02-22
docker-compose version 2.1 (v2.1)


vito-c:
I just changed the version from:
```
version: '2'
services:
```
to
```
version: '2.1'
services:
```
and  I get the following error:
```
ERRO[0000] Could not parse config for project libcompose : yaml: unmarshal errors:
  line 1: cannot unmarshal !!str `2.1` into config.RawService
panic: yaml: unmarshal errors:
  line 1: cannot unmarshal !!str `2.1` into config.RawService
```

Are there plans to support version 2.1 in the future? There are some important features added to the compose file `healthcheck` and `depends_on` was improved.

vdemeester:
@vito-c yes there is 👼 Just lagging behind quite a bit (running a little bit out of time 😅)

vito-c:
I would love to help is there a branch? or does it need to be implemented from scratch?

vito-c:
@vdemeester anything I can do to help?

joshwget:
It sounds like the changes to `depends_on` aren't supported in v3 anymore (https://github.com/docker/compose/blob/master/CHANGELOG.md#compose-file-version-21-and-up). Is it worth working on this feature still?

vito-c:
@joshwget backwards compatibility is important.

I'm also a little confused by the decision to drop support for `depends_on` and `health_checks` I haven't really heard a good argument for doing so.

yelvert:
+1

cheapRoc:
I've [started working](https://github.com/cheapRoc/libcompose/commits/refactor-v2_1) on some of this in my spare time. I'm not 100% positive why the changes were made in [this commit](https://github.com/docker/libcompose/commit/99e58bed93c224d4cc7f49a1a643e5854094155f) but I've got past the general YAML parsing error mentioned in this issue simply by removing some unused portion of that commit (there's a double YAML parse within `CreateConfig`).

In order to begin the support of more compose file versions (#474, #490) I hope to continue refactoring between when the compose file is parsed and when it is validated by JSON schema. I think that should at least bump this up to '2.1'. 

I'll try to post something up early to see what the original authors think. Cheers.

joyel24:
+1 for support 2.1 yml

mlosev:
I think we also need 2.2 schema support

donbobka:
+1 version 2.3

lpalgarvio:
https://github.com/rancher/rancher/issues/10158

CpuID:
[...]

PR #497, opened 2017-10-12
Parse, validate, and load docker-compose v2.1 files

cheapRoc:
Ref: #441 

The goal of this PR is to hopefully introduce v2.1 formatted compose files so I can use `libcompose` for various stuff I'm working on around Joyent's Elastic Docker Host (Triton).

I've started by refactoring how compose files are loaded and validated by their version numbers. This solves the immediate issue of #441 by cleaning up how the version is referenced when parsing compose YAML. At the moment, I've left off cleaning up how JSON schema files are loaded and passed into `config/validation.go`. This is working well so I'll start utilizing this for loading v2.1 specifically.

I'm working on this in my spare time ATM (which isn't much) so bare with me. Also, I'm opening this up to get any early feedback. I do realize I need tests for some stuff. 😄

### DIFF
--- a/config/merge.go
+++ b/config/merge.go
@@ -9,6 +9,7 @@ import (
 
 	"reflect"
 
+	"github.com/Sirupsen/logrus"
 	"github.com/docker/docker/pkg/urlutil"
 	"github.com/docker/libcompose/utils"
 	composeYaml "github.com/docker/libcompose/yaml"
@@ -110,17 +111,14 @@ func Merge(existingServices *ServiceConfigs, environmentLookup EnvironmentLookup
 		}
 	}
 
+	parts := strings.Split(config.Version, ".")
+	major, _ := strconv.Atoi(parts[0])
 	var serviceConfigs map[string]*ServiceConfig
 
-	switch config.Version {
-	case "2.1":
-		// STUB until we can properly validate and parse minor versions
-		var err error
-		serviceConfigs, err = MergeServicesV2(existingServices, environmentLookup, resourceLookup, file, baseRawServices, options)
-		if err != nil {
-			return "", nil, nil, nil, err
-		}
-	case "2":
+	switch major {
+	case 3:
+		logrus.Fatal("Note: Compose file version 3 is not yet implemented")
+	case 2:
 		var err error
 		serviceConfigs, err = MergeServicesV2(existingServices, environmentLookup, resourceLookup, file, baseRawServices, options)
 		if err != nil {
--- a/config/merge.go
+++ b/config/merge.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"reflect"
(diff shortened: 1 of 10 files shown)


cheapRoc:
I see the lints and probably borked tests... I'll touch up on that next time.

config/merge.go
```diff
@@ -25,14 +27,21 @@ var (
...
+// format validation should be handled by the caller.
 func CreateConfig(bytes []byte) (*Config, error) {
 	var config Config
 	if err := yaml.Unmarshal(bytes, &config); err != nil {
 		return nil, err
 	}
 
-	if config.Version != "2" {
+	parts := strings.Split(config.Version, ".")
+	major, _ := strconv.Atoi(parts[0])
```
vdemeester: We shouldn't skip the error here

config/merge.go
```diff
@@ -102,14 +111,20 @@ func Merge(existingServices *ServiceConfigs, environmentLookup EnvironmentLookup
 		}
 	}
 
+	parts := strings.Split(config.Version, ".")
+	major, _ := strconv.Atoi(parts[0])
```
vdemeester: Same here (error skipping)

config/schema.go
```diff
@@ -7,479 +7,922 @@ var schemaDataV1 = `{
   "type": "object",
 
   "patternProperties": {
-    "^[a-zA-Z0-9._-]+$": {
-      "$ref": "#/definitions/service"
-    }
+	"^[a-zA-Z0-9._-]+$": {
```
vdemeester: I think we should extract those in `json` files and have `go generate` create this file from it (like we do on `docker/cli`). Not sure if we should do it in the  PR or not

config/merge.go
cheapRoc: Normally I'm 100% with you, and please correct me if I'm wrong, but I'm not sure what error we'd capture here or would want to bubble up considering that we're guaranteed to have `0` output by [`strconv.Atoi`](https://github.com/golang/go/blob/46aa9f5437b000fad3696b0cd9fd97995da16411/src/strconv/atoi.go#L202) regardless of input or error. We should always handle a `0` versioned compose file since they're the default in all cases.

config/schema.go
cheapRoc: I agree, I was hoping someone would mention this. It's definitely annoying to have them in a huge file.

cheapRoc: I checked out `docker/cli` ([this is it correct?](https://github.com/docker/cli/tree/master/cli/compose/schema)) and noticed that it uses `go-bindata`. I've used that in the past for storing JSON schemas into a binary loading them at run time, so I understand the context. Is that how you foresee doing things in `libcompose`?

If so, I can probably swing that.

config/merge.go
cheapRoc: @vdemeester, interested in your thoughts here. Thanks.

config/schema.go
vdemeester: @cheapRoc yep :angel:

config/merge.go
vdemeester: @cheapRoc `version` is of type string (in the schema), so I could put `a.b` it would pass the validation. That's why I think we should check for errors.

cheapRoc: I agree, we should validate the full version number since it's a string. Maybe that's something that needs to be handled before this spot (?). That's not what I was trying to achieve here.

At this line, and the other you pointed out, I'm only parsing a single integer value. The `a` of `a.b`, and I should be handling the default case as `0` appropriately.

cheapRoc: That’s what I’ll do, I’ll formally parse the schema version major and minor and try to reuse it in other places.

cheapRoc:
This has fallen off my immediate road map as the project I intended to use it for hasn't materialized. If I have the free time I'll re-open the PR. Apologies!

Issue #494, opened 2017-09-18
compose Up fails trying to recreate containers when not needed


lucacome:
I can't seem to find a way to replicate the behavior of `docker-compose up -d`, where if the container has not changed, nothing happens. Instead, I get:
`ERRO[0006] Failed to start: rethinkdb : Error response from daemon: endpoint with name rethinkdb already exists in network ubuntu_default`

Am I doing something wrong?

lucacome:
I'm guessing it's because of this set to true: https://github.com/docker/libcompose/blob/master/docker/service/service.go#L242

vito-c:
I can confirm this was introduced by this commit: 
https://github.com/docker/libcompose/commit/f5c34b0a35164a0f6d8b861df32b2e4e54deb847

vito-c:
@xihan88

vito-c:
Simple Expected Example with docker compose:
```yaml
---
version: '2'
services:
    test.svr:
        container_name: test.svr
        image: alpine:latest
        command: sleep 10000
```
```
$ docker-compose up -d
Creating network "foo_default" with the default driver
Creating test.svr ...
Creating test.svr ... done
vito-mbp:maestro vitocutten$ docker-compose up -d
test.svr is up-to-date
```
Then make a change to the compose file `ports ["8080:8080"]` for example
```
$ docker-compose up -d
Recreating test.svr ...
Recreating test.svr ... done
```

Libcompose is no longer following this behavior it is now barfing out an error on second up.

lucacome:
Is anything happening about this?

xihan88:
@lucacome This issue is due to the problem from L331-L347 in this file https://github.com/docker/libcompose/blob/57bd716502dcbe1799f026148016022b0f3b989c/docker/service/service.go#L331
because if a container is not recreated, it shouldn't be connecting to the previous network and starting again. I will make a simple fix for it.
@vito-c In the previous version before my PR, a service will always be recreated (because of hashing an array of addresses) hence this issue won't show itself. I'm glad it brings out this important bug.

PR #504, opened 2018-01-24
Fix: no connect & start if not recreated

xihan88:
Signed-off-by: xi.han <xi.han@sihuatech.com>
This is a simple way to fix https://github.com/docker/libcompose/issues/494 . However I believe there are better ways for this issue. For example **recreateIfNeeded()** could return a bool showing whether the service was updated or not.


xihan88:
@vdemeester Could you help me to trigger the test again? Thanks!

PR #512, opened 2018-09-14
Extra options for "build" sections in YAML

CpuID:
Adds support under the `build` heading for:

- `cache_from`
- `labels`
- `network`
- `target`

Note: `network` is undocumented in the [docker-compose.yml reference](https://docs.docker.com/compose/compose-file/) but is supported and works (as of config version 3.4).

@vdemeester do we also need these in [docker/service/service.go](https://github.com/docker/libcompose/blob/master/docker/service/service.go#L176-L186)?

### DIFF
--- a/config/schema.go
+++ b/config/schema.go
@@ -223,7 +223,7 @@ var servicesSchemaDataV2 = `{
                 "cache_from": {"$ref": "#/definitions/list_of_strings"},
                 "labels": {"$ref": "#/definitions/list_or_dict"},
                 "network": {"type": "string"},
-                "target": {"type": "string"},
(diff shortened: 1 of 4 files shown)


CpuID:
Now that's interesting, the acceptance tests fail yet Jenkins passes?

vdemeester:
@CpuID yes, the acceptance tests is merely there to see how close (or not) we are to support the full range of `docker-compose` feature, that's why it's not taken into account in the final result of the build… Otherwise, no tests would pass… :sweat_smile:

CpuID:
np thx, calling "mission accomplished" in that case :)

